PR #6, opened 2023-05-20
Initial proof-of-concept plugin implementation

codefromthecrypt:
#### What type of PR is this?

/kind feature

This adds an implementation that requires an SDK. For example, the guest must export the function "filter".

Note: Go won't work with this until v1.22 because it doesn't yet support `//go:wasmexport`, even if TinyGo works today.

n/a

This code is in proof-of-concept phase as merging will allow more people to participate. Shortly after, we should setup CI to ensure it continues to compile and run.

Conversion is still required. For example, we are manually choosing fields to convert from `v1.PodSpec` because its `v1.PodSpec.Marshal` isn't compatible with: `protoapi.IoK8SApiCoreV1PodSpec.UnMarshalVT`

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilter
BenchmarkFilter/success:_node_is_match_with_spec.NodeName
BenchmarkFilter/success:_node_is_match_with_spec.NodeName-12         	   92158	     12846 ns/op
PASS
```

n/a as unreleased


codefromthecrypt:
ps while this doesn't actually do any protobuf decoding (for reasons mentioned in the PR of help wanted), it does work and has a base overhead of below, which includes time to initialize the module in tinygo.

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilter
BenchmarkFilter/success:_node_is_match_with_spec.NodeName
BenchmarkFilter/success:_node_is_match_with_spec.NodeName-12         	   29073	     45341 ns/op
PASS
```

In this use case, tinygo is an order of magnitude faster than the WIP Go 1.21. This is expected as it is the first run. I only am testing this to show it works.
```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilter
BenchmarkFilter/success:_node_is_match_with_spec.NodeName
BenchmarkFilter/success:_node_is_match_with_spec.NodeName-12         	     294	   4098859 ns/op
PASS
```

codefromthecrypt:
I added a commit to show protobuf on each side, but it is using the unreleased GOOS=wasip1 as tinygo (even dev) doesn't currently compile. If someone knows how to avoid the issues or help raise to tinygo much appreciated as the performance with GOOS=wasip1 isn't good.

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilter
BenchmarkFilter/success:_node_is_match_with_spec.NodeName
BenchmarkFilter/success:_node_is_match_with_spec.NodeName-12         	     103	  10949407 ns/op
PASS
```

sanposhiho:
PTAL at https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/7 whether it'll help you or not.
(I'll push the commit to this branch if you want, instead of the separate PR)

codefromthecrypt:
I added commits from #7 thanks. What's left is the following before we can proceed to ABI:

1. plugin/testdata make never responds on my host. I wonder if it is taking a very long to compile or something else. If someone can try, it can verify if tinygo actuaully works now.
2. In plugin/vfs/plugin.go I have the below, though I don't know that the v1.NodeSpec type isn't already in a format decodable by the protos generated from tinygo. The problem is since I can't complete the guest I can't test it. If we do need to convert first, I hope we can do so automatically because the types are very big.

```go
	case "pod/spec":
		// TODO msg := convert(*v1.PodSpec)
		var msg protoapi.IoK8SApiCoreV1PodSpec
		marshaller = func() ([]byte, error) {
			return proto.Marshal(&msg)
		}
```

appreciate a hand if anyone can confirm the above

codefromthecrypt:
ok current status: on @evacchi's advice, I added in github.com/planetscale/vtprotobuf/cmd/protoc-gen-go-vtproto to generate the `UnmarshalVT` functions. Sadly, things still don't compile (in TinyGo)

TinyGo hangs compiling possibly due to https://github.com/tinygo-org/tinygo/issues/3653#issuecomment-1529847344

plugin/vfs/plugin.go
```diff
@@ -0,0 +1,127 @@
...
+	pod      *v1.Pod
+	nodeInfo *framework.NodeInfo
+}
+
+func (v vfs) Open(name string) (fs.File, error) {
+	var marshaller func() ([]byte, error)
+	switch name {
+	case "pod/spec":
+		// TODO v.pod.Spec.Marshal is incompatible, find a way to automatically
+		// convert *v1.PodSpec to protoapi.IoK8SApiCoreV1PodSpec
```
codefromthecrypt, line 66: this part is still an issue, we at least need a better comment as to why these are different encoding (e.g. what encoding `v1.PodSpec.Marshal` is and why it isn't the same as `protoapi.IoK8SApiCoreV1PodSpec.MarshalVT`

codefromthecrypt:
I switched from using planetscale/vtprotobuf to using @knqyf263's https://github.com/knqyf263/go-plugin to generate the protos (despite no service being defined). This got around the tinygo hang in compilation, so we can proceed!

codefromthecrypt:
I have the proof-of-concept working now. It has 2 approaches VFS and ABI. I updated the description about the two. Here is an example run of benchmarks between the two:

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilterABI
BenchmarkFilterABI/success:_node_is_match_with_spec.NodeName
BenchmarkFilterABI/success:_node_is_match_with_spec.NodeName-12         	  176182	      6659 ns/op	     184 B/op	       6 allocs/op
BenchmarkFilterVFS
BenchmarkFilterVFS/success:_node_is_match_with_spec.NodeName
BenchmarkFilterVFS/success:_node_is_match_with_spec.NodeName-12         	   27487	     43850 ns/op	  156675 B/op	      88 allocs/op
PASS
```

kerthcet:
Thanks @codefromthecrypt I'll take a look today.

sanposhiho:
@codefromthecrypt 
I haven't looked into the impl, but has this been solved by something then?
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/6#discussion_r1199854277

codefromthecrypt:
@sanposhiho 

> I haven't looked into the impl, but has this been solved by something then?
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/6#discussion_r1199854277

Nothing has changed on this point. I still want to know why the types aren't compatible. I'm using a similar approach as your wasm POC and just converting an arbitrary field to make it pass. I hope at some point, even after merge, how to address conversion is addressed

guest/abi/internal/imports/imports.go
```diff
@@ -0,0 +1,15 @@
...
+
+package imports
+
+//go:wasm-module k8s.io/scheduler
+//go:export reason
+func _reason(ptr, size uint32)
+
+//go:wasm-module k8s.io/api
+//go:export nodeInfo/node/name
+func _nodeInfoNodeName(ptr uint32, limit bufLimit) (len uint32)
```
codefromthecrypt, line 11: In both implementations, I am choosing nodeInfo/node/name and pod/spec just to make the demo work. I don't expect this to be the final form of the plugin. The main thing is to defer unmarshalling until requested. So a complete impl may just have 2 functions `nodeInfo` and `pod`. The nice thing about having a base case is being able to compare before/after decisions (because benchmark are already here).

codefromthecrypt:
Also, I want to note that I will improve the unit test coverage, but don't want to do this until the PR main code is accepted. Otherwise it can be a lot of work just to throw away.

codefromthecrypt:
update: so I spent even more hours trying to find any solution to external generation of Go source that can match what's done in k8s.io/api/core/v1.

TL;DR; This should be raised to its own issue, as data type conversion is a small part of the work, though required for any serious use.

The problem is our Makefile generates wire-incompatible source, so the types are only useful as value types, not as types for marshaling. For example, the k8s library type `ServiceAccount` type has different wire type numbers than our `IoK8sApiCoreV1ServiceAccount`.

I also looked at possibly using [json instead](https://github.com/kubewarden/go-policy-template/blob/cb700974909e544a7823199b8db184723abe7a6f/validate.go#L37-L38), but this brings in a different set of compilation problems, as well the performance of converting back and forth from json.

The above commentary is a more elaborated version of [this comment](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/6#discussion_r1199854277) due to hours spent trying to get it to work or find any single project that has had success without re-using k8s libraries or json.

codefromthecrypt:
ps if we have to manually convert, I would highly suggest moving to [karmem](https://github.com/inkeliz/karmem), which was designed for fast wasm.

Doing exactly the same as this PR, the "abi" guest wasm reduces to 8KB (from 330) and the request time, even without pooling, is <1us vs 6.67us

I suspect even if we flesh out the model, it will still be a lot more efficient.

```bash
$ cat kubernetes/karmem/kubernetes.km 
karmem api;

struct Node table {
	Name []char;
}

struct NodeInfo table {
	Node Node;
}

struct PodSpec table {
	NodeName []char;
}

struct Pod table {
	Spec PodSpec;
}
```

Then using the objects built with `karmem build --golang -o api kubernetes.km`

guest decoding change
```diff
-func (pod) Spec() *protoapi.IoK8SApiCoreV1PodSpec {
+func (pod) Spec() *karmemapi.PodSpec {
        b := imports.PodSpec()
-       var msg protoapi.IoK8SApiCoreV1PodSpec
-       if err := msg.UnmarshalVT(b); err != nil {
-               panic(err)
-       }
-       return &msg
+       podSpec := new(karmemapi.PodSpec)
+       podSpec.ReadAsRoot(karmem.NewReader(b))
+       return podSpec
 }
```

host encoding change
```diff
-               // TODO v.pod.Spec.Marshal is incompatible, find a way to automatically
-               // convert *v1.PodSpec to protoapi.IoK8SApiCoreV1PodSpec
-               var msg protoapi.IoK8SApiCoreV1PodSpec
-               msg.NodeName = v.pod.Spec.NodeName
-               marshaller = msg.MarshalVT
+               marshaller = func() ([]byte, error) {
+                       writer := karmem.NewWriter(1024)
+
+                       content := karmemapi.PodSpec{
+                               NodeName: v.pod.Spec.NodeName,
+                       }
+
+                       if _, err := content.WriteAsRoot(writer); err != nil {
+                               return nil, err
+                       }
+                       return writer.Bytes(), nil
+               }
```

again, this doesn't even do any pooling of readers or writers.

sanposhiho:
One thing I mind about Karmem is the limitation of the languages. If we decide to use Karmem, doesn’t it mean that the guest should be implemented in the languages which Karmem supports?

plugin/abi/plugin.go
```diff
@@ -0,0 +1,245 @@
...
+	buf := uint32(stack[0])
+	bufLimit := bufLimit(stack[1])
+
+	nodeInfoNodeName := filterArgsFromContext(ctx).nodeInfo.Node().Name
+	nodeInfoNodeNameLen := writeStringIfUnderLimit(mod.Memory(), buf, bufLimit, nodeInfoNodeName)
+
+	stack[0] = uint64(nodeInfoNodeNameLen)
+}
+
+func apiPodSpec(ctx context.Context, mod wazeroapi.Module, stack []uint64) {
```
sanposhiho, line 181: @codefromthecrypt 

_(not sure it’s worth mentioning now since it’s PoC / we can do such optimization later as a follow-up)_

I’m wondering that... can we somehow re-use allocated object across multiple scheduling cycles instead of allocating every time `apiPodSpec` is called as long as the object isn't updated on the cluster? 
Or such "reuse the address to the same place" is stupidly dangerous and should be avoided?
 
My expectation is that:
- have a goroutine that is working on replacing the allocated object only when that resource is updated on the cluster. 
- during scheduling, from the plugin to guest, we just give a reference to an object to a guest.

codefromthecrypt:
@sanposhiho TL;DR; don't worry about karmem for the first round. The real problem is the size of the model, as even in karmem when I make the model complete, it takes a long time per request.

---

I have an update as I manually converted the entire model to Karmem locally. I wanted to do a fair comparison. Basically, there is not much time difference for what we are doing now. The huge model is the problem in the current benchmark, much more than the specific marshaller. Compared to the results in the PR desc, you can see the baseline overhead is less, but not a huge mount less than proto. It is smaller (96kb), but that also may be due to stripping all the rpc code out of it.

```
goos: darwin
goarch: arm64
pkg: sigs.k8s.io/kube-scheduler-wasm-extension/plugin
BenchmarkFilterABI
BenchmarkFilterABI/success:_node_is_match_with_spec.NodeName
BenchmarkFilterABI/success:_node_is_match_with_spec.NodeName-12         	  183386	      6450 ns/op
BenchmarkFilterVFS
BenchmarkFilterVFS/success:_node_is_match_with_spec.NodeName
BenchmarkFilterVFS/success:_node_is_match_with_spec.NodeName-12         	   26743	     44040 ns/op
PASS
```

plugin/abi/plugin.go
codefromthecrypt: yeah I would like to do this part later. There are issues because the guest is garbage collected.

We can't write to the same place especially as we might write more than originally allocated. It would be better to not mess with the guest memory, IMHO. On the other hand, we can optimize serialization of the object host-side so we are just copying the byte slice over (instead of re-serializing it). However, I think we can drive this with benchmarks, and we can optimize this at any time.

codefromthecrypt:
Suggestions after spending many hours on pros and cons of serialization. Main thing is I think we should move ahead and not get too caught in the weeds for the first code. Especially, there are many areas of optimization (host side, guest side etc), as well conversion and modeling work, but these are only viable once an initial impl is in.

* Let's at least tentatively accept the overhead of proto. We may still end up a lot faster than we have to be.
* Let's kick conversion or wire type matching to a new issue, because this does need to resolve, but not for POC.
* Let's get feedback about what should change here for our first cut. We can iterate after. For example, tell me if you want me to remove "vfs" to make it more simple.

kerthcet:
Hey guys, I have to confess I'm not quite familiar with wasm tech right now, but I'm learning hard and trying my best to understand it by reading the docs and designs and practice them locally. My question is is this PR the minimalist one, it's quite large and hard to review, or that's how wasm works? And can we have some docs or do we have any plan, future readers will benefit from this as well. 

My thought is we can have a minimalism version and let's merge it ASAP, then we can build more things upon it. Does this make sense? 

Anyway, really impressed works from @codefromthecrypt I'm actually learning something from your videos from youtube.

kerthcet:
Hey guys, I have to confess I'm not quite familiar with wasm tech right now, but I'm learning hard and trying my best to understand it by reading the docs and designs and practice them locally. My question is is this PR the minimalist one, it's quite large and hard to review, or that's how wasm works? And can we have some docs or do we have any plan, future readers will benefit from this as well. 

My thought is we can have a minimalism version and let's merge it ASAP, then we can build more things upon it. Does this make sense? 

Anyway, really impressed works from @codefromthecrypt I'm actually learning something from your videos from youtube.

sanposhiho:
@codefromthecrypt 

(I forgot to put the comment here as well) I'm OK to remove VFS stuff from here. We can recall VFS idea in the future if we need to.

>  I think we should move ahead and not get too caught in the weeds for the first code.

Also, I agree with this. Instead of stopping here, it's better to proceed with the current only option Karmem. 
We can think more on this after some iteration since the time hopefully solves something on this. (via TinyGo's new version, or Golang official wasi support etc)

> I have to confess I'm not quite familiar with wasm tech right now

@kerthcet 

No worries, me neither 🙃 
For me, reading http-wasm implementation helped me a lot to know the overview of what we want to do in this project, as an example.
https://http-wasm.io/http-handler-abi/

> My question is is this PR the minimalist one, it's quite large and hard to review, or that's how wasm works? And can we have some docs or do we have any plan, future readers will benefit from this as well.

I agree about the doc. I'd be great if this PR gets some brief doc + example impl of the guest. @codefromthecrypt 
but I'd say they don't need to be a perfect one for now; the implementation is, either way, kind of PoC/draft even after merging, and we can grow them up as we grow the implementation.

Regarding the amount of the code, this PR gets smaller after VFS stuff removal (+ probably proto stuff as well?) 
And after that, it looks reasonable to me to review/merge it without splitting into smaller PRs anymore; I'd say that splitting PRs for each of guest/host doesn't help us make it easier to read because we anyway want to read both implementations at the same time.

> My thought is we can have a minimalism version and let's merge it ASAP, then we can build more things upon it. Does this make sense?

👍 
We may come up with further improvement on this like my rough idea above, we can pile up such optimization later

kerthcet:
And about the `kubernetes.proto`, seems we container the whole native CRDs, am I right? And we met the problems with ServiceAccount as @codefromthecrypt mentioned above (have no idea whether this is resolved or not), but we can start with a subset of the CRDs, because for scheduler that's pretty too much. I can think of the necessary API objects, like:
- Pod
- Node
- PV & PVC
- Namespace
- ConfigMap

Can we clip the proto file, where we get this? @sanposhiho

codefromthecrypt:
I will pare this down to the minimum ABI. PS I plan to continue to use proto, not Karmem, for the time being. Both work, and we can swap proto for Karmem later if we like. I think the best way to start though is with familiar tooling, even if a little less efficient.

example/main.go
```diff
@@ -0,0 +1,18 @@
+package main
+
+import "sigs.k8s.io/kube-scheduler-wasm-extension/guest"
+
+func main() {
+	guest.FilterFn = filter
+}
+
+func filter(nodeInfo guest.NodeInfo, pod guest.Pod) (guest.Code, string) {
```
codefromthecrypt, line 9: here is the example

plugin/host.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	filterArgsFromContext(ctx).reason = reason
+}
+
+func k8sApiNodeInfoNodeFn(ctx context.Context, mod wazeroapi.Module, stack []uint64) {
+	buf := uint32(stack[0])
+	bufLimit := bufLimit(stack[1])
+
+	node := filterArgsFromContext(ctx).nodeInfo.Node()
+	// TODO: fields are different between v1.Node and V1Node
```
codefromthecrypt, line 80: here are some type-mapping concerns I think we should address after this PR is merged.

codefromthecrypt (commented):
ok I've simplified the implementation. If this is decent shape for future work, let me know and I will backfill tests for merge.

kerthcet:
> I will pare this down to the minimum ABI. PS I plan to continue to use proto, not Karmem, for the time being. Both work, and we can swap proto for Karmem later if we like. I think the best way to start though is with familiar tooling, even if a little less efficient.

Agree, we should also consider codebase dependency problems. Just in case some of the project will not maintain in the future.

codefromthecrypt:
thanks @kerthcet. I assume we are on the same page, so I'll backfill tests etc and ping back for final review.

plugin/host.go
sanposhiho: 👍

example/main.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+)
+
+func main() {
+	guest.Filter = api.FilterFunc(nameEqualsPosSpec)
+}
+
+// nameEqualsPosSpec schedules this node if its name equals its pod spec.
```
sanposhiho, line 12: We can come back to such interface topic later, but at final, we probably want to make the interface of wasm guest very similar to the existing scheduling framework interface. 
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/interface.go#L373

guest/go.mod
```diff
@@ -0,0 +1,13 @@
+module sigs.k8s.io/kube-scheduler-wasm-extension/guest
+
+go 1.19
```
sanposhiho, line 3: Can we use 1.20 like other modules?

plugin/host.go
```diff
@@ -0,0 +1,97 @@
...
+// k8sSchedulerStatusReasonFn is a function used by the wasm guest to set the
+// framework.Status reason.
+func k8sSchedulerStatusReasonFn(ctx context.Context, mod wazeroapi.Module, stack []uint64) {
+	ptr := uint32(stack[0])
+	size := bufLimit(stack[1])
+
+	var reason string
+	if b, ok := mod.Memory().Read(ptr, size); !ok {
+		// don't panic if we can't read the message.
+		reason = "BUG: out of memory reading message"
```
sanposhiho, line 68: Just curious, what happen if the host function panices?

sanposhiho (approved):
Thanks @codefromthecrypt for the awesome work :)

There are some things we need to discuss/improve based on here, but, this implementation is a really good starting point for us!
Regarding the problem around serialization, I'd also say it's OK to skip via a small conversion (from k8s struct to our struct) for now. Let's revisit it later.

Rather than continuing the discussion on this PR for a long time, I'd like to merge this and move each discussion topic to issues.

/lgtm

I'd leave approval to @kerthcet.

sanposhiho:
/unhold
/remove-approval

OK, learn that the approver's approval is regarded as `/approval`. 😅

kerthcet:
It's merged.... But at least we have a base codeline.

sanposhiho:
@kerthcet 
Sorry, my `/hold` cannot stop him from merging this. (Seems I should have run `/remove-approve`)
Please take a look from your side, and give some feedback!

@codefromthecrypt please check my small comments above!

plugin/host.go
sanposhiho: Leave my curiosity aside. 
In such error cases, we want to return an error from `Filter()` to the scheduling framework eventually.

plugin/plugin.go
```diff
@@ -0,0 +1,99 @@
...
+		instanceCounter:   atomic.Uint64{},
+	}
+
+	// Eagerly add one instance to the pool. Doing so helps to fail fast.
+	g, err := pl.getOrCreateGuest(ctx)
+	if err != nil {
+		_ = runtime.Close(ctx)
+		return nil, err
+	}
+	pl.pool.Put(g)
```
sanposhiho, line 44: Cannot we initialize one guest here and use the same guest instance in all functions? What's the reason of the need of pool?

codefromthecrypt:
don't worry I can help address any comments post merge. I will focus on them soon. Thanks for moving forward together!

guest/go.mod
codefromthecrypt: I think we should switch this when tinygo 0.28 is out cc @deadprogram https://github.com/tinygo-org/tinygo/milestone/14

plugin/host.go
codefromthecrypt: so any host function panic is returned as an error from wazero api.Function call. So this means the scheduling framework sees it as an error return.

plugin/plugin.go
codefromthecrypt: my assumption is that there are multiple parallel requests to the scheduler. If not, we can indeed make it a single guest.

example/main.go
codefromthecrypt: right I was thinking that `api.Filter` is similar to that, excluding the ctx parameter, and temporarily excluding `state` as I don't have a clear idea what that will map to.

plugin/plugin.go
sanposhiho: Got it, thanks!
Yes, each plugin can receive multiple parallel requests. So, I think that makes sense.

sanposhiho: fyi @pohly

guest/go.mod
sanposhiho: Ah, limitation via tinygo side. OK, that makes sense.

example/main.go
sanposhiho: Ah, yes. a bit different topic though, we definitely need to support cycle state somehow eventually.
The cycle state is the object which is initialized at the beginning of each scheduling. And we use it as a way to pass something from one extension point to another. One common usecase is pre-calculate something in `PreFilter` or `PreScore` and use pre-calculation result in `Filter` or `Score`. (PreFilter/PreScore is called once per one scheduling, but Filter/Score is called for every potential Nodes. That's why we want to do precalculation instead of calculating every time Filter/Score is called)
https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/#extension-points

codefromthecrypt: good. I think we can stub in a state param and develop it further. just like now we stubbed in the other params but they are not fully working yet due to conversion issues

plugin/plugin.go
pohly: > no wasm guest is safe for concurrent use

Is that a limitation of the current runtime and may that get addressed at some point?

codefromthecrypt: TL;DR; I think at least for a year maybe two, assume all guests have to be used sequentially.

Even if the wasm runtime supports atomics, the guest would have to be compiled to use them. One issue is garbage collector implementations (compiled to wasm) would need to use them. It isn't just that, as stack pointers etc would also need to be changed. I don't expect TinyGo or Go to do that for a long time. Even though there's a GC proposal in wasm which may affect some of this, it wasn't written with Go in mind. For example, it doesn't support interior pointers. I don't think any garbage collected language compiled to wasm, or even a malloc/free language is using atomics yet.

codefromthecrypt:
I raised this to track switching to the same protos as the scheduler plugin uses. We'd still use go-plugin to compile it, just a different source, so that the fields and wire types match https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/13

plugin/plugin.go
kerthcet: > Yes, each plugin can receive multiple parallel requests

Can you elaborate? Why would we receive parallel requests?

sanposhiho: Regarding the scheduling cycle, yes. But we have a binding cycle running in parallel.

sanposhiho: Also, we run Filter plugins in default preemption as well
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/preemption/preemption.go#LL606C77-L606C77

example/main.go
sanposhiho: I expect that the difficulty to support the cycle state is that people can insert any data in the cycle state.
We don't know what they're, and the cycle state also contains the mutex, which probably difficult to pass.

Probably two options here?
- don't support the cycle state referred from other plugins.
  - = we don't need to pass the cycle state from guest to host; we can just keep it in the guest side.
- require people to define how to serialize their cycle state data. And the guest serialize the data based on that to pass it to host.
  - take a lock via the host function?

sanposhiho: Anyway created two followup issues to discuss them.

- #15
- #16

pohly: > we don't need to pass the cycle state from guest to host; we can just keep it in the guest side.

That won't work because currently, different guest instances may be invoked for the same cycle - see https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/6#discussion_r1204239244

Then guest A sets up its cycle state, but guest B gets called later and doesn't have it.

sanposhiho: Right... But, we may be able to give the same instance for the same Pod's scheduling.
Looking at each scheduling, the plugin is not called in parallel except [during the preemption](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/preemption/preemption.go#LL606C77-L606C77) (I need to have a double check if there are no other places to call same plugin in parallel)

We can take alock or something during the preemption. Yes, then the performance problem may be coming next though.

PR #7, opened 2023-05-21
add protobuf definition and generated Golang code which is buildable with TinyGo

sanposhiho:
#### What type of PR is this?

/kind feature

add protobuf definition + generated Golang protobuf code which is buildable with TinyGo.
We cannot use the proto definitions in kubernetes/api because it depends on many stuff which aren't buildable with TinyGo.
(Hopefully, it'll be buildable in the future version of TinyGo.)

The protobuf definition is generated from https://github.com/nytimes/openapi2proto with [the patch](https://github.com/sanposhiho/openapi2proto/pull/1). And use `google.golang.org/protobuf/cmd/protoc-gen-go` for Golang code generation.

As you see in ./kubernetes/proto/go.mod, it only depends on `google.golang.org/protobuf`, which is buildable with TinyGo.

Fixes #

```release-note
NONE
```


codefromthecrypt:
Maybe you can add a test that shows hot to do the following flow? This will help make sure it still works. The test may need to be in the plugin module, in order to avoid leaking v1 deps to the guest.

* Convert from `v1.Pod` to `api.IoK8SApiCoreV1Pod`
* Marshal `api.IoK8SApiCoreV1Pod` to proto
* UnMarshal `api.IoK8SApiCoreV1Pod` from proto
* Convert back to `v1.Pod` (optional)

codefromthecrypt (approved):
Feel free to merge and/or push commits to my branch.

codefromthecrypt:
I made notes on #6 and not sure how much of https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/7#issuecomment-1556129351 is needed. Basically, first thing is to make sure we can build tinygo wasm which reads it. For some reason my laptop doesn't respond when trying to build, both 0.27 and dev version of tinygo.

codefromthecrypt:
In #6 I added code to generate the MarshallVT functions

sanposhiho:
It's included in #6.

/close

Issue #9, opened 2023-05-24
setup scheduler with the wasm extension


sanposhiho:
**What would you like to be added**:

```[tasklist]
### Tasks
- [x] create the scheduler with the wasm extension
- [x] create `registry.k8s.io/kube-scheduler-wasm-extension/scheduler` to provide the scheduler in which the wasm extension is enabled
```

**Why is this needed**:

We will eventually try to include this wasm extension into the upstream, but it's worth having the scheduler binary/image here meanwhile regardless.

I'll work on this since I recently worked on the registry stuff in the simulator and understand the process.

/assign
/cc @kerthcet

sanposhiho:
I meant building the scheduler with our [wasm extension plugin](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/plugin/plugin.go)
Just like scheduler-plugins does: https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/cmd/scheduler/main.go

kerthcet:
Got it, plz go ahead.

Gekko0114:
/lgtm
Thanks!

sanposhiho:
Still there is a remaining step for the registry.

PR #18, opened 2023-05-28
feature: set up the scheduler with wasm plugin

sanposhiho:
#### What type of PR is this?

/kind feature

- Set up the scheduler with the wasm plugin.
  - Move the wasm plugin directory under `/scheduler/pkg/plugins`.
- Implement the plugin config for the wasm plugin.
  - Need this to register the wasm plugin in `app.WithPlugin`

Part of #9 

```release-note
- Set up the scheduler with the wasm plugin
- The plugin config for the wasm plugin is implemented.
```


sanposhiho:
cc @codefromthecrypt

scheduler/pkg/plugins/wasm/config.go
```diff
@@ -0,0 +1,8 @@
+package wasm
+
+type PluginConfig struct {
+	// GuestName is the name of the guest wasm.
+	GuestName string `json:"guestName"`
+	// GuestPath is the path to the guest wasm.
+	GuestPath string `json:"guestPath"`
```
codefromthecrypt, line 7: cool. later I will help switch this to URI based

codefromthecrypt (approved):
thanks for the help!

scheduler/pkg/plugins/wasm/config.go
kerthcet: FYI, maybe GuestName might be optional as we can parse the .wasm file. Anyway, would not block this PR.

kerthcet (commented):
/lgtm
/approve
/hold

I think this would introduce code conflicts, unhold if you think it's ok.

scheduler/pkg/plugins/wasm/guest.go
```diff
@@ -1,4 +1,4 @@
-package plugin
```
kerthcet, line 1: Nit: folder `scheduler` might not be necessary because this is for kube-scheduler.

sanposhiho:
/unhold

Yes, I meant to do this one.

scheduler/pkg/plugins/wasm/guest.go
sanposhiho: I'd like to keep the scheduler folder so that we can separate the multiple go modules in serial

```
./guest: guest wasm Go module
./scheduler: host of wasm guest. (actually, the plugin in it is the guest though)
```

codefromthecrypt:
> FYI, maybe GuestName might be optional as we can parse the .wasm file. Anyway, would not block this PR.

agree it is optional even if we parse the wasm file as only thing it does is improve logs

Issue #16, opened 2023-05-24
support the cycle state in guest


sanposhiho:
**What would you like to be added**:

Support [CycleState](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/cycle_state.go#L48) in guest.

**Why is this needed**:

The cycle state is one of the important struct passed to each extension point, which is initialized at the beginning of each scheduling. 
And we use it as a way to pass something from one extension point to another. 
One common usecase is pre-calculate something in PreFilter or PreScore and use pre-calculation result in Filter or Score. (PreFilter/PreScore is called once per one scheduling, but Filter/Score is called for every potential Nodes. That's why we want to do precalculation instead of calculating every time Filter/Score is called)
https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/#extension-points

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

sanposhiho:
>  think with some thinking, we can pin a guest to cycle state if that helps make things more efficient or less complex.

I have a similar thought.
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/6#discussion_r1207996051

The preemption is the only place where the same plugin is called for the same scheduling cycle in parallel. (talking with my memory, need to double-check) So, in other places, we should be able to basically use the same instance for the same scheduling cycle.

codefromthecrypt:
ok so if I understand, there are two concerns:
* the same memory/instance should be used for the whole scheduling cycle
* even in the same scheduling cycle, the instance may be called in parallel.

So, I think that to solve this, we need to:
* pin an instance to the scheduling cycle. (take a module instance from pool on `PreFilter`, release back on `Permit`)
* within the scheduling cycle, even using the same instance, make sure there's exclusive access to it (lock on the module instance).

Is this correct?

sanposhiho:
Yes, correct. That is mostly the same as what I'm thinking.

> pin an instance to the scheduling cycle. (take a module instance from pool on PreFilter, release back on Permit)

But, Permit may not be called actually, for example, when all Nodes are filtered out in Filter phase. 
So, we can simplify it by taking an instance from pool on PreFilter and then if PreFilter is called with a different Pod, we can assume the past scheduling cycle was finished and release back an instance.

sanposhiho:
[...]

PR #19, opened 2023-05-28
feature(guest): implement framework interface which is quite similar to the scheduling framework

### LINKED ISSUE #15: Make the guest interface very similar to existing scheduler plugins interface
**What would you like to be added**:

people can implement wasm based plugins in very similar interface of the [plugins](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/interface.go#L303-L505) in scheduling framework.

**Why is this needed**:

You can see how to implement the guest in the [example](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/example/main.go).
It's now quite different from how we implement the scheduler plugins.

To reduce the learning cost for people and to make it easy to convert from existing plugins to wasm-based plugins, we want to reduce the difference from [the scheduler's plugin interface](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/interface.go#L303-L505) as much as possible.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

sanposhiho:
#### What type of PR is this?

/kind feature

This PR defines the interface which is very similar to the scheduling framework and allows users to implement the guest plugin in a similar way to what they do in native Go scheduler plugins.

Fixes #15

```release-note
Improve the guest interface exposed to users so that users can implement the guest plugin in a similar way to what they do in native Go scheduler plugins.
```


guest/api/framework/interface.go
```diff
@@ -0,0 +1,178 @@
...
+	// the given node fits the pod. If Filter doesn't return "Success",
+	// it will return "Unschedulable", "UnschedulableAndUnresolvable" or "Error".
+	// For the node being evaluated, Filter plugins should look at the passed
+	// nodeInfo reference for this particular node's information (e.g., pods
+	// considered to be running on the node) instead of looking it up in the
+	// NodeInfoSnapshot because we don't guarantee that they will be the same.
+	// For example, during preemption, we may pass a copy of the original
+	// nodeInfo object that has some pods removed from it to evaluate the
+	// possibility of preempting them to schedule the target pod.
+	Filter(ctx context.Context, state *CycleState, pod PodFn, nodeInfo NodeInfoFn) *Status
```
sanposhiho, line 36: One difference from the scheduler's interface:
For lazy fetching, I give here PodFn and NodeInfoFn instead of actual objects.

codefromthecrypt: ok just if CycleState is big, we end up lazy again, and then the issue becomes back to lazy all things. We can find out.

codefromthecrypt, line 36: I don't think we should add ctx parameter here as it won't be what people think it is. The ctx will have no value in tinygo as it isn't related to the host context.

codefromthecrypt (approved):
This works, we can reset our benchmark expectations to this design and work from there. Thanks for getting the things aligned.

sanposhiho:
/unhold

Thanks! @kerthcet

guest/api/framework/interface.go
sanposhiho: let's continue the effort on https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/16. :)

sanposhiho: Well. I'd like to keep it to make the interface as similar to the scheduler's one as possible even if we cannot pass the context from host to guest. But, as you said, not to make people confused, we need to note somewhere about that at least though.

Also, making it similar to the scheduler's original interface is not only the reason. This PR [just passes](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/1da62a0ca248d800be114ef343b0576be9e98a69/guest/filter.go#L23) the plain `context.Background()` to user-defined Filter, but in the future, we may want to do something further like:
- have logger in the ctx so that people can use the logger from ctx with [FromContext](https://github.com/sanposhiho/kubernetes/blob/8badc6f8fe45e2133ca23ba00ecba94dbe04c8d8/vendor/k8s.io/klog/v2/contextual.go#L156)
  - In Kubernetes, it's very common to create logger from ctx.
- `calcel()` passed context somewhere when the guest instance is going to be deleted.

Those things could be reasons we pass `ctx` to users, what do you think?

codefromthecrypt: So this is in the guest code, and it cannot inherit context from the host. So, let's contain this discussion to the sandbox.

Since the guest can't inherit the context, the only thing that would be passed here is a global variable. Since no callback can return a modified context, that variable would never change. So, a context parameter (which has cost to have in the parameter list) will be no different than a field named context in practice.

example/main.go
```diff
@@ -17,23 +17,39 @@
...
+
+func (*NodeName) Name() string {
+	return name
 }
 
-// nameEqualsPodSpec schedules this node if its name equals its pod spec.
-func nameEqualsPodSpec(args api.FilterArgs) (api.StatusCode, string) {
-	nodeName := nilToEmpty(args.NodeInfo().Node().Metadata.Name)
-	podSpecNodeName := nilToEmpty(args.Pod().Spec.NodeName)
+func (*NodeName) Filter(ctx context.Context, state *framework.CycleState, podFn framework.PodFn, nodeInfoFn framework.NodeInfoFn) *framework.Status {
```
codefromthecrypt, line 45: This implementation shouldn't need a pointer receiver.

example/main.go
```diff
@@ -17,23 +17,39 @@
...
-	guest.Filter = api.FilterFunc(nameEqualsPodSpec)
+	guest.RegisterFilter(&NodeName{})
+}
+
+// NodeName is a plugin that checks if a pod spec node name matches the current node.
+type NodeName struct{}
+
+const (
+	// Name is the name of the plugin used in the plugin registry and configurations.
+	name = "NodeName"
```
codefromthecrypt, line 35: I think having these as constants only used in one place each adds indirection which reduces readability of the example, which is otherwise simple. I suggest inlining them.

example/main.go
```diff
@@ -17,23 +17,39 @@
...
-func nameEqualsPodSpec(args api.FilterArgs) (api.StatusCode, string) {
-	nodeName := nilToEmpty(args.NodeInfo().Node().Metadata.Name)
-	podSpecNodeName := nilToEmpty(args.Pod().Spec.NodeName)
+func (*NodeName) Filter(ctx context.Context, state *framework.CycleState, podFn framework.PodFn, nodeInfoFn framework.NodeInfoFn) *framework.Status {
+	nodeName := nilToEmpty(nodeInfoFn().Node().Metadata.Name)
+	podSpecNodeName := nilToEmpty(podFn().Spec.NodeName)
 
 	if len(podSpecNodeName) == 0 || podSpecNodeName == nodeName {
-		return api.StatusCodeSuccess, ""
+		return nil
```
codefromthecrypt, line 50: I think we are trying too hard to be exactly the same as the host code, and this won't be the same. Not only is the context parameter questionable, but so is this, We are adding the question of "why and when should we return nil?" vs passing the two components.

codefromthecrypt (approved):
I have to say that I don't think this change makes things better, and it would be slower and allocate more while also adding confusion. However, I am not terribly strong in this opinion so go through with it if you like.

I think the plugin is different in wasm, and especially when you look at other languages, the subtle differences are worth knowing by signature instead of docs about why the signature doesn't match practical concerns of the guest.

guest/api/framework/interface.go
codefromthecrypt: It is helpful to remember that we can't do anything you can't do in another language like C. There's no special thing about the host written in Go, that allows the guest to do anything go-specific. Not sure if this thinking helps, but maybe!

example/main.go
sanposhiho: In the scheduler plugins, nil means success.
https://github.com/kubernetes/kubernetes/blob/bdf34b3f56d269d9b435a882145aa7fe217691ce/pkg/scheduler/framework/interface.go#L78

example/main.go
sanposhiho: What's the reason in your mind?

sanposhiho:
> it would be slower and allocate more 

I wouldn't say allocating `framework.Status` isn't an expensive cost, it's reasonable to take it and make the interface the same as the scheduler's one.

> while also adding confusion

If you mean nil / pointer staff which you commented, they're not confusing to people usually working around the scheduler plugins.

example/main.go
codefromthecrypt: In general, pointer receivers are made to modify the object they refer to. there's no value to refer to, so it doesn't make sense (to me).

example/main.go
codefromthecrypt: I'm fine with going conventionally. Just remember this plugin will eventually support other languages and also not necessarily be slow because proto reasons. We can proceed to optimize for conventionality instead of performance, just remember performance can degrade quickly and some people will fault the plugin after it reaches some threshold. If we ignore performance over a long time it can be difficult to recover.

codefromthecrypt:
>  it's reasonable to take it and make the interface the same as the scheduler's one.

I think eventually you will decide that the interface won't be the same as the host side, but now you don't see this. I may be wrong on this, so just defer my comments until things like what I said about context make sense to you. It is better that we proceed than get hung up on things that can be learned later.

guest/api/framework/interface.go
sanposhiho: Yes, I know we cannot pass context from the host. But, I mean we can pass something (like listed in [it](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/19#discussion_r1209353906)) to the user-defined `Filter` function [here](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/1da62a0ca248d800be114ef343b0576be9e98a69/guest/filter.go#L23).
It includes the design choice as well, @kerthcet what is your take on this?

codefromthecrypt: I don't think we should encourage this sort of thinking inside the guest. using context as a substitute for a global logger, or doing things like guest-specific deadlines are going to add overhead and ease people's misunderstanding of Wasm-specific concerns especially horrible performance vs native, no concurrency etc.

sanposhiho:
It's quite related to the user experience. Regarding not only this `Filter` function, but all functions in all extension points, I'd like to keep the original scheduling framework's interface as much as possible. (As I do `pod PodFn, nodeInfo NodeInfoFn`, I don't disagree with the interface change that must need)

I know it makes sense that it's faster to return raw values instead of returning them in one fancy struct (framework.Status).
But, on the other hand, we've already had many future _potential_ wasm-extension users in the world, those using the native Golang scheduler plugins. And, making many unnecessary differences between our interface and the scheduling framework will directly be the migration costs for users. The plugin tends to be invoked in several extension points.

guest/api/framework/interface.go
codefromthecrypt: here's the only example I can find of context being used in tinygo. Remember this is single threaded. Basically, main wires up this function and creates the root context which is a cancel context. There's no way to import an incoming context from the host.

https://github.com/fastly/compute-sdk-go/blob/main/fsthttp/handle.go#L17-L40

sanposhiho: Wait, allocating something on the guest vs allocating the same thing on the host; is there a performance difference between them? If so, that'll completely change my mind.

sanposhiho: Also, can you elaborate on "no concurrency"? Cannot user use goroutine in guest?

codefromthecrypt: yes there is a huge difference, this is why I raised the PR about benchmarks showing that guest take 10x as long as the host on the same proto. garbage collection is inefficient and there's only one thread. Even if we did add context as a parameter we would really want to test that it can do what we think it can.

The main thing is you can't "lift and shift" code without thinking into webassembly today for reasons like this. I would prefer to make small changes, and evaluate them with tests. Many loggers for example aren't really usable in wasm, not practically anyway. We have to unfortunately be very deliberate if we want the users' code to succeed.

My 2p.

codefromthecrypt: so again I'm ok with doing a big change, but I think we will learn more from smaller ones.

For example, it is one change to add different hooks, and another to have those hooks return pointers to values, and another to add context support (actually wired up not just mocked). If we do these things separately, and seems we have time to, we can learn one by one impact, and more importantly it is easier to write tests.

codefromthecrypt: > Also, can you elaborate on "no concurrency"? Cannot user use goroutine in guest?

So you can launch goroutines, but all execution happens on the calling thread. This means things like GC are a part of execution time https://wazero.io/languages/tinygo/#concurrency

Once the benchmarks PR merges, I'll add profiling so you can see what I mean with our example code.

sanposhiho: 🤯 OK, that's my missed ones.

Then, I'd like to abandon this PR for now. Not saying we don't aim to make it similar to the original one, but on the road toward there, we should pile up smaller changes one by one while checking the performance.

sanposhiho:
/close

https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/19#discussion_r1210982138

guest/api/framework/interface.go
codefromthecrypt: I think having a goal to be like the current SDK is shared, how we get there, and how much we can practically speaking are the differences. What I like about this PR is it shows how it might look. Thanks for listening!

PR #21, opened 2023-05-28
feature: always assign the same instance to the same Pod

sanposhiho:
#### What type of PR is this?

/kind feature

This PR changes the host to always assign the same guest instance to the same Pod.
Two motivations here.

##### 1. support cycle state #16 

This change simplifies the problem how to support the cycle state. That's difficult because users can put any data into cycle state, and we don't know what'll be put.
After this PR, we don't need to pass the cycle state between host and guest; the guest initializes the cycle state inside of it and uses the same cycle state in the same scheduling.

Obviously, this idea means that users cannot pass the data from one plugin to other plugin through the cycle state, but that usecase is rare. (Actually, AFAIK, no in-tree plugins do that) 
Not saying we never support it in the future though, I'd say we don't need to prioritize it.

##### 2. cache resource data

Thanks to the snapshot, the NodeInfo and Pod passed to each extension point aren't get changed during the same scheduling cycle. We don't need to pass them to guest every time we call the guest. The guest can cache them during the same scheduling cycle.

Part of #16 

Detailed description: https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/16#issuecomment-1566352602

```release-note
NONE
```

### DIFF
--- a/scheduler/go.mod
+++ b/scheduler/go.mod
@@ -38,7 +38,8 @@ require (
 	github.com/tetratelabs/wazero v1.1.1-0.20230520044102-8c7e0caead29
 	k8s.io/api v0.26.2
 	k8s.io/apimachinery v0.26.2
-	k8s.io/kubectl v0.0.0
+	k8s.io/component-base v0.26.2
+	k8s.io/kubectl v0.26.2
 	k8s.io/kubernetes v1.26.2
 )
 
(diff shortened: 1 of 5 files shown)


sanposhiho:
cc @codefromthecrypt

codefromthecrypt (commented):
I think we need to get benchmarks in and do before/after for changes like this as some things may end up being more expensive than before.

What I would guess needs to happen on the guest, is that the current pod needs to be a global (even if not exported) field.

The guest will call to get the pod, and the host tracks if it has been sent or not. We need a special value return to say the guest shouldn't try to re-write that field based on if it is invalidated or not.

I think it will looks something like this:

```go
var cachedPod *protoapi.Pod
var cachedPodCookie uint32

func (filterArgs) Pod() *protoapi.Pod {
	b, newCookie := imports.Pod(cachedPodCookie)
	if newCookie != cachedPodCookie {
		if err := cachedPod.UnmarshalVT(b); err != nil {
			panic(err.Error())
		}
		cachedPodCookie = newCookie
	}
	return cachedPod
}
```

The main change to the ABI is accepting a cookie and packing the "new cookie" into the result with the length.

When the input cookie is zero, that's the base case, first time the pod is initialized on the guest. Later calls when the host sees the cookie is current, it doesn't write anything to memory at all, just returns the same cookie value as the input. The guest helper code notices this and doesn't update anything.

codefromthecrypt (approved):
I'm cool with how this is going, and made a suggestion of what I expect the guest side to need to do in order to avoid redoing unmarshal

kerthcet:
> pass the data from one plugin to other plugin

I heard of such case in the community meeting for performance, but I can't recall it. I think it's ok because all the plugins up-to-now are independent, that's why we need plugin.

sanposhiho:
just rebased.

sanposhiho:
> pass the data from one plugin to other plugin

Actually my company also does that actually. 😅 But, yes we could say that's very minor usecase that we can give a lower priority.

kerthcet:
> Actually my company also does that actually.

What's your scenario?

sanposhiho:
> What's your scenario?

We have a plugin very similar to [coscheduling](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/coscheduling/README.md). And we want to notify the post filter plugin (which is also a custom one.) if the Pod during scheduling is the first Pod among the pod group.

scheduler/pkg/plugins/wasm/plugin.go
```diff
@@ -67,25 +158,44 @@ func (pl *wasmPlugin) Name() string {
...
 		}
 	}
-	return poolG.(*guest), nil
+
+	return poolG, nil
+}
+
+func (pl *wasmPlugin) PreFilter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod) (*framework.PreFilterResult, *framework.Status) {
+	// When PreFilter is called, run unassign because the previous scheduling cycle should have been finished.
+	pl.pool.unassign()
```
kerthcet, line 178: Can we move `unassign` to PostFilter and Reserve? Because then we can cover all the unassign scenarios.
- scheduling cycle filter & score error --> PostFilter
- scheduling cycle reserve & permit error --> Unreserve
- binding cycle error --> Unreserve
- scheduling & binding success --> PostBind

sanposhiho: No, PostFilter won't be executed when the scheduling cycle is failed in extension points other than PreFilter and Filter.
So, the PreFilter is the only place where we can notice that the past scheduling cycle was finished and another scheduling cycle gets started.

codefromthecrypt (approved):
When this is in, I'll help spike a way to avoid decoding the same pod many times. It is easier to show than tell.

scheduler/pkg/plugins/wasm/plugin.go
kerthcet: Yes, recheck the logic, we still have one gap here when assuming the pod. 

Then another question is when podA is in async binding, podB get scheduled and ran into PreFilter, it will unassign the `assignedToSchedulingPod`, so that means the instance for scheduling cycle and binding cycle might be different, as expected? What's the advantage vs _using one instance across scheduling & binding_? cc @codefromthecrypt

sanposhiho: In that case, before PodA proceeding to binding cycle, Permit should move the instance assigned to PodA from assignedToSchedulingPod to assignedToBindingPod. So, The instance for scheduling cycle of PodA and binding cycle of PodA should be the same.

kerthcet: Oh that's the logic for `assignForBinding`

codefromthecrypt:
> Thanks, https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/21#pullrequestreview-1448527569 right? Sorry forgot to reply, but cookie design looks very good.

yes except maybe I will call it etag not cookie. In any case we need to extract or derive 32-bits from the UUID.

scheduler/pkg/plugins/wasm/plugin.go
kerthcet: Then no need to distinguish with scheduling guest or binding guest, we can use the podUID as the identifier, and adding a new field like `schedulingGuests` to contain all the in use guest instances. wdyt?  Anything I missed here?

kerthcet: > we cannot know which guest instance that we need to take lock during preemption

Doesn't the instance related to the podUID? Can't we get the instance from the `schedulingGuests` via UID? Why the preemption has something to do with the binding cycle?  Still confused.

kerthcet: > a summary should be in the code comments

Can't agree more.

sanposhiho: Added the note on the `guestPool`.

sanposhiho:
rebase + have a change for feedback from @kerthcet

sanposhiho:
rebased

codefromthecrypt (approved):
I'd like to get this in as it lets me build on it, for caching node data in the guest between calls.

PR #30, opened 2023-06-06
ci: adds basic setup

codefromthecrypt:
#### What type of PR is this?

/kind cleanup

This sets up basic CI with a couple notable targets:

* `make check`: runs various linters and formatters, and ensures projects build.
* `make test-scheduler`: runs scheduler-based unit and e2e tests.

None

Later, we can make tests that run in tinygo for the guest, but as that's tricky I left things here for now.

NONE

N/A

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -1,0 +1,63 @@
+# `name` value will appear "as is" in the badge.
+# See https://docs.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository
+# yamllint --format github .github/workflows/commit.yaml
+---
+name: "build"
+
+on:
+  push:  # We run tests on non-tagged pushes to main
+    tags: ''
+    branches: main
+    paths-ignore:
+      - '**/*.md'
+  pull_request:  # We also run tests on pull requests targeted at the main branch.
+    branches: main
+    paths-ignore:
+      - '**/*.md'
+  # workflow_dispatch will let us manually trigger the workflow from GitHub actions dashboard.
+  # For example, you can try to build a branch without raising a pull request.
+  # See https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow
+  workflow_dispatch:
+
+jobs:
+  check:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        go-version:  # Note: Go only supports 2 versions: https://go.dev/doc/devel/release#policy
+          - "1.20"
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
+          cache: false  # cache separately to include golangci-lint
+
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/.cache/golangci-lint
+            ~/go/pkg/mod
+            ~/go/bin
+          key: check-${{ runner.os }}-go-${{ matrix.go-version }}-${{ hashFiles('**/go.sum', 'Makefile') }}
+
+      - run: make check
+
+  test-scheduler:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        go-version:  # Note: Go only supports 2 versions: https://go.dev/doc/devel/release#policy
+          - "1.20"
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - run: make test-scheduler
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -7,6 +11,18 @@ build-tinygo: examples/filter-simple/main.wasm examples/noop/main.wasm
 %/main-debug.wasm: %/main.go
 	@(cd $(@D); tinygo build -o main-debug.wasm -scheduler=none -target=wasi main.go)
 
+.PHONY: testdata
+testdata:
+	@$(MAKE) build-tinygo
+	@$(MAKE) build-wat
+
+.PHONY: build-wat
+build-wat: $(wildcard scheduler/test/testdata/*/*.wat)
```
codefromthecrypt, line 20: this makefile change will conflict with #25, but I will handle it later. To me it doesn't matter which is merged first, but we should try to merge one or the other (or both) today because the build is currently broken.

codefromthecrypt:
rebased. I'm not sure if actions are disabled on this repo, but most of the change only helps if they are.

codefromthecrypt:
rebased. main thing left in here is lint fix, fix of tinygo go.mod, and basic CI plumbing

Makefile
```diff
@@ -72,3 +76,59 @@ update-kubernetes-proto: proto-tools
...
+test-scheduler:
+	@for d in ./scheduler ./internal/e2e; do \
+        (cd $$d; go test ./...); \
+	done
+
+.PHONY: check  # Pre-flight check for pull requests
+check:
+	@$(MAKE) tidy
+	@$(MAKE) build
+	@$(MAKE) lint golangci_lint_goarch=arm64
```
anuraaga, line 128: I guess there isn't much of a chance of having amd/arm-specific code in this repo like wazero, so does it make sense to just run lint once with default settings, using `go run`? I do sometimes have issues in the wazero repo because of version drift of the installed golangci binary, but don't know of any alternative there because of the arch-specific code. If it's not an issue in this repo, `go run` may be better

codefromthecrypt: good point only once makes sense

codefromthecrypt: A lot better, thanks for calling attention to it!

Makefile
```diff
@@ -72,3 +76,51 @@ update-kubernetes-proto: proto-tools
 		--go-plugin_opt=Mk8s.io/apimachinery/pkg/runtime/generated.proto=sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/runtime \
 		--go-plugin_opt=Mk8s.io/apimachinery/pkg/runtime/schema/generated.proto=sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/schema \
 		--go-plugin_opt=Mk8s.io/apimachinery/pkg/util/intstr/generated.proto=sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/instr;
+	@$(MAKE) format
+
+.PHONY: lint
+lint: $(golangci_lint_path)
```
anuraaga, line 82:
```suggestion
lint:
```

.github/workflows/commit.yaml
```diff
@@ -0,0 +1,62 @@
...
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        go-version:  # Note: Go only supports 2 versions: https://go.dev/doc/devel/release#policy
+          - "1.20"
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-go@v3
```
anuraaga, line 33: Looks like using `v4` elsewhere but not here, if it's because of the default cache behavior, maybe can use `v4` and `cache: false`

Makefile
```diff
@@ -72,3 +76,51 @@ update-kubernetes-proto: proto-tools
...
+	@for d in ./scheduler ./internal/e2e; do \
+        (cd $$d; go test ./...); \
+	done
+
+.PHONY: check  # Pre-flight check for pull requests
+check:
+	@$(MAKE) tidy
+	@$(MAKE) build
+	@$(MAKE) lint
+	@$(MAKE) format
```
anuraaga, line 122: Nit, you may move format up so easy formatting misses fail / can be fixed faster

anuraaga (approved):
Thanks!

.github/workflows/commit.yaml
codefromthecrypt: accident

codefromthecrypt: well you are right ;) missing comment which should be there with the false

Makefile
codefromthecrypt: good idea

codefromthecrypt:
thanks @anuraaga I wove in the feedback and preemptively squashed as usually asked to and doing that loses approvals.

kerthcet (commented):
/lgtm
approve

Let's see how it works.

codefromthecrypt:
Thanks for the help getting this together @anuraaga and thanks for being open to trying non-standard CI @kerthcet!

PR #31, opened 2023-06-06
extracts out pool and stubs plugin functions

codefromthecrypt:
#### What type of PR is this?

/kind cleanup

This extracts the guest pool into its own type so that it can be unit tested without export helpers. Notably, this makes it a generic type which makes testing a lot easier. In the process, I found and fixed a test bug.

This also stubs all the host-side plugins in declaration order according to the framework interface. This helps me understand what we are implementing, and allows the IDE to find the upstream documentation vs guessing.

https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/interface.go#L304-L506

NONE

NONE

N/A

### DIFF
--- a/scheduler/plugin/host.go
+++ b/scheduler/plugin/host.go
@@ -19,11 +19,10 @@ package wasm
 import (
 	"context"
 
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/kubernetes/pkg/scheduler/framework"
-
 	"github.com/tetratelabs/wazero"
 	wazeroapi "github.com/tetratelabs/wazero/api"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
 )
 
 const (
@@ -54,11 +53,11 @@ func instantiateHostApi(ctx context.Context, runtime wazero.Runtime) (wazeroapi.
 		Instantiate(ctx)
 }
 
-// filterArgsKey is a context.Context value associated with a filterArgs
+// filterParamsKey is a context.Context value associated with a filterParams
 // pointer to the current request.
-type filterArgsKey struct{}
+type filterParamsKey struct{}
 
-type filterArgs struct {
+type filterParams struct {
 	pod      *v1.Pod
 	nodeInfo *framework.NodeInfo
 	// reason is a field to avoid compiler-specific malloc/free functions, and
@@ -67,8 +66,8 @@ type filterArgs struct {
 	reason string
 }
 
-func filterArgsFromContext(ctx context.Context) *filterArgs {
-	return ctx.Value(filterArgsKey{}).(*filterArgs)
+func filterParamsFromContext(ctx context.Context) *filterParams {
+	return ctx.Value(filterParamsKey{}).(*filterParams)
 }
 
 // k8sSchedulerStatusReasonFn is a function used by the wasm guest to set the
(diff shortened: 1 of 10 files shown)


scheduler/plugin/plugin.go
```diff
@@ -20,17 +20,14 @@ import (
...
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	frameworkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
-
-	"github.com/tetratelabs/wazero"
 )
 
 const PluginName = "wasm"
```
codefromthecrypt, line 33: all logic in this file is the same, I only implemented the plugin interfaces, stubbed with TODO.

scheduler/plugin/pool.go
```diff
@@ -0,0 +1,142 @@
...
+	"context"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// guestPool manages guest to pod assignments in a scheduling or binding cycle.
+type guestPool[guest any] struct {
+	// newGuest is a function to create a new guest.
+	newGuest func(context.Context) (guest, error)
```
codefromthecrypt, line 29: By extracting this function and making it a generic type, I was able to find a subtle bug in the test.

scheduler/plugin/plugin_test.go
```diff
@@ -26,57 +26,13 @@ import (
...
-	}
-	defer p.(io.Closer).Close()
-
-	pl, ok := wasm.NewTestWasmPlugin(p)
-	if !ok {
-		t.Fatalf("failed to cast plugin to wasmPlugin: %v", ok)
-	}
-
-	uid := types.UID("test-uid")
-	differentuid := types.UID("test-uid")
```
codefromthecrypt, line 48: this was the bug in test which drove me a little bonkers until I saw it ;)

codefromthecrypt (commented):
notes

scheduler/plugin/pool.go
```diff
@@ -0,0 +1,142 @@
...
+func (p *guestPool[guest]) assignForBinding(podUID types.UID) {
+	g := p.assignedToSchedulingPod
+	p.assignedToBindingPod[podUID] = g
+}
+
+// unassignForScheduling unassigns the guest instance from the pod puts it back
+// into the pool.
+func (p *guestPool[guest]) unassignForScheduling() {
+	assigned := p.assignedToSchedulingPod
+	var zero guest
```
codefromthecrypt, line 107: only weird thing for people not used to generics is you can't cast to nil (`guest(nil)`), However the following has the same effect.

kerthcet (commented):
Only one nit, other LGTM

scheduler/plugin/host.go
```diff
@@ -67,8 +66,8 @@ type filterArgs struct {
 	reason string
 }
 
-func filterArgsFromContext(ctx context.Context) *filterArgs {
-	return ctx.Value(filterArgsKey{}).(*filterArgs)
+func filterArgsFromContext(ctx context.Context) *filterParams {
```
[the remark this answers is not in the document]
codefromthecrypt: done and squashed. thanks for catching this!

kerthcet:
/lgtm
/approve

thanks

codefromthecrypt:
thanks for the review @kerthcet!

PR #32, opened 2023-06-07
makes guest API more similar to host

codefromthecrypt:
#### What type of PR is this?

/kind cleanup
/kind api-change

This makes the parameters to filter function more like those in the host by creating a proxy interface. Doing so only slows down the no-op case by some tens of nanoseconds, so it doesn't matter.

The interesting part is if users return nil for status instead of multiple values it is faster. So, this changes signature to return a pointer to status. (I checked against stack based value and it is still faster).

Finally, this optimizes the example, because our test data doesn't actually have podSpecNodeName field. This drop in execution time proves parameters not read are lazy.

NONE because the project isn't usable yet.

First, changing to new guest signatures
```
goos: darwin
goarch: arm64
                                            │   old.txt   │              new.txt               │
                                            │   sec/op    │   sec/op     vs base               │
PluginFilter/noop/params:_small-12            147.2n ± 1%   187.8n ± 0%  +27.54% (p=0.002 n=6)
PluginFilter/noop/params:_real-12             148.4n ± 1%   186.4n ± 1%  +25.57% (p=0.002 n=6)
PluginFilter/filter-simple/params:_small-12   7.274µ ± 1%   6.518µ ± 6%  -10.39% (p=0.002 n=6)
PluginFilter/filter-simple/params:_real-12    113.9µ ± 1%   101.2µ ± 1%  -11.21% (p=0.002 n=6)
geomean                                       2.063µ        2.192µ        +6.25%
```

Next, optimizing the filter example
```
goos: darwin
goarch: arm64
                                            │   old.txt    │              new.txt               │
                                            │    sec/op    │   sec/op     vs base               │
goos: darwin
goarch: arm64
                                            │   old.txt    │              new.txt               │
                                            │    sec/op    │   sec/op     vs base               │
PluginFilter/noop/params:_small-12             147.2n ± 1%   187.3n ± 2%  +27.20% (p=0.002 n=6)
PluginFilter/noop/params:_real-12              148.4n ± 1%   187.1n ± 3%  +26.08% (p=0.002 n=6)
PluginFilter/filter-simple/params:_small-12    7.274µ ± 1%   6.564µ ± 3%   -9.75% (p=0.002 n=6)
PluginFilter/filter-simple/params:_real-12    113.95µ ± 1%   49.28µ ± 2%  -56.75% (p=0.002 n=6)
geomean                                        2.063µ        1.835µ       -11.05%
```


Makefile
```diff
@@ -5,7 +5,7 @@
 build-tinygo: examples/filter-simple/main.wasm examples/noop/main.wasm
 
 %/main-debug.wasm: %/main.go
-	@(cd $(@D); tinygo build -o main-debug.wasm -scheduler=none -target=wasi main.go)
+	@(cd $(@D); tinygo build -o main-debug.wasm -gc=custom -tags=custommalloc -scheduler=none -target=wasi main.go)
```
codefromthecrypt, line 8: build was broke

codefromthecrypt:
ps I ran tests manually, and they pass. after https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/30 we don't need manual verification

guest/filter.go
```diff
@@ -24,57 +24,90 @@ import (
...
+func (n *nodeInfo) Node() *protoapi.Node {
+	return n.node()
 }
 
-func (filterArgs) Pod() *protoapi.Pod {
-	b := imports.Pod()
-	var msg protoapi.Pod
+func (n *nodeInfo) node() (node *protoapi.Node) {
+	if node = n.n; node != nil {
+		return
```
anuraaga, line 72: Minor but I think using non-naked return makes this block easier to grok at the cost of just 5 letters

guest/filter.go
```diff
@@ -24,57 +24,90 @@ import (
...
+}
+
+func (p *pod) Status() *protoapi.PodStatus {
+	return p.pod().Status
+}
+
+// pod lazy initializes p from the imported host function imports.Pod.
+func (p *pod) pod() (pod *protoapi.Pod) {
+	if pod = p.p; pod != nil {
+		return
```
anuraaga, line 104: Ditto

anuraaga (commented):
As may be noticed by now I've started watching this repo and figured can leave some comments. I can stop though if it's annoying 😄

sanposhiho (commented):
one nit

codefromthecrypt:
@anuraaga your review is helpful thanks! many things I'm not committed to, and it is helpful to see someone looking!

guest/filter.go
codefromthecrypt: agree!

codefromthecrypt:
I should disclose the simple wasm filter grew 9KB over this, but it is 9/876 and probably unanimously worth it

guest/internal/imports/host.go
```diff
@@ -16,10 +16,30 @@
...
-import "runtime"
+import (
+	"runtime"
+
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+)
+
+// StatusToCode returns a WebAssembly compatible result for the input status,
+// after sending any reason to the host.
+func StatusToCode(s *api.Status) uint32 {
```
codefromthecrypt, line 27: refactored this out so I can use it in `ScorePlugin` without copy/pasta

codefromthecrypt:
@anuraaga can you take another look and approve or let me know if any glitches left? since this has no formal approvals to lose yet, I snuck in some more godoc and polish.

guest/filter.go
```diff
@@ -24,57 +24,87 @@ import (
...
+func (p *pod) Spec() *protoapi.PodSpec {
+	return p.pod().Spec
+}
+
+func (p *pod) Status() *protoapi.PodStatus {
+	return p.pod().Status
+}
+
+// pod lazy initializes p from the imported host function imports.Pod.
+func (p *pod) pod() *protoapi.Pod {
```
codefromthecrypt, line 98: later I want to make this generic, but not sure you can make a generic function of a pointer...

anuraaga (approved):
Thanks!

guest/filter/filter.go
```diff
@@ -0,0 +1,50 @@
...
+// For example:
+//
+//	func main() {
+//		filter.Plugin = api.FilterFunc(nameEqualsPodSpec)
+//	}
+var Plugin api.FilterPlugin
+
+// filter is only exported to the host.
+//
+//go:export filter
```
codefromthecrypt, line 38: putting this in a separate package allows the guest developer to opt-into what it is using in a way the host can see.

scheduler/plugin/guest.go
```diff
@@ -98,3 +94,26 @@ func decorateError(out fmt.Stringer, fn string, err error) error {
...
+
+const (
+	exportFilterPlugin exports = 1 << iota
+)
+
+func detectExports(exportedFns map[string]wazeroapi.FunctionDefinition) (exports, error) {
+	var e exports
+	for name, f := range exportedFns {
+		switch name {
+		case guestExportFilter:
```
codefromthecrypt, line 108: as we add new plugins, we add detection logic here. Some plugins might be composite because the developer wants to implement both filter and score for better performance vs defining multiple plugins.

scheduler/plugin/plugin.go
```diff
@@ -41,12 +42,28 @@ func New(configuration runtime.Object, frameworkHandle framework.Handle) (framew
...
+	return maskInterfaces(plugin), err
+}
+
+// maskInterfaces ensures the caller can do type checking to detect what the
+// plugin supports.
+func maskInterfaces(plugin *wasmPlugin) framework.Plugin {
+	if plugin == nil {
+		return nil
+	}
+	if plugin.guestExports == exportFilterPlugin {
```
codefromthecrypt, line 55: This is how we support current day plugin use, where a type assertion happens on the result of a pluginFactory.

as we support more plugin types, this logic will get more complicated, a cartesian product perhaps of many combinations. However, the cardinality of plugin types is fixed.

scheduler/plugin/plugin_test.go
```diff
@@ -170,9 +194,14 @@ func TestNewFromConfig(t *testing.T) {
 			name:      "valid wasm",
 			guestPath: test.PathExampleFilterSimple,
 		},
+		{
+			name:          "not plugin",
+			guestPath:     test.PathErrorNotPlugin,
+			expectedError: `wasm: guest does not export any plugin functions`,
```
codefromthecrypt, line 200: now we show how something that is wasm, but not a plugin shows up. Basically it is an error not export support for at least one plugin type.

codefromthecrypt (commented):
I revised for work-in-progress score code, since no approvals yet. Notably I found out that the plugin system today relies on type assertions and we didn't handle that. Now we do.

scheduler/plugin/plugin.go
```diff
@@ -57,6 +74,14 @@ func NewFromConfig(ctx context.Context, config PluginConfig) (framework.Plugin,
 		return nil, err
 	}
 
+	var guestExports exports
+	if guestExports, err = detectExports(guestModule.ExportedFunctions()); err != nil {
+		_ = runtime.Close(ctx)
+		return nil, err
+	} else if guestExports == 0 {
```
kerthcet, line 81: Shouldn't we close runtime here?

kerthcet (commented):
one question, then I can approve, cc @sanposhiho for more advices.

scheduler/plugin/plugin.go
codefromthecrypt: yep

codefromthecrypt: there was actually another, too. I refactored the code to make leaks less easy.

sanposhiho (approved):
/lgtm

several good ideas here, thanks

PR #33, opened 2023-06-09
have a label tide/merge-method-squash in all PRs

sanposhiho:
#### What type of PR is this?

/kind cleanup

have a label `tide/merge-method-squash` in all PRs via PR template.
Instead of asking squash for authors, we can ask the bot to do squash merge.

Fixes #

```release-note
NONE
```

```benchstat

```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -25,6 +25,8 @@ Optionally add one or more of the following kinds if applicable:
 /kind regression
 -->
 
+/label tide/merge-method-squash
+
 #### What this PR does / why we need it:
 
 #### Which issue(s) this PR fixes:
@@ -36,6 +38,7 @@ _If PR is about `failing-tests or flakes`, please post the related issues/tests 
 Fixes #
 
 #### Special notes for your reviewer:
+
 
 #### Does this PR introduce a user-facing change?
 <!--


codefromthecrypt (approved):
does this limit the message to only the title of PR? GitHub has this feature now, and it is a lot nicer that a list of "broke, fixed, fuzz" etc.

anuraaga:
If the bot is using the GitHub API it probably reflects the repo settings, so it may be good to confirm what they look like. The setting is around here
image

sanposhiho:
> does this limit the message to only the title of PR? GitHub has this feature now, and it is a lot nicer that a list of "broke, fixed, fuzz" etc.

It also has the commits like https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/commit/f801a888e637e897ddca822ae213780e30a89030

sanposhiho:
I just changed the setting to allow only squash merging, let's see how the bot works in another PR

Screenshot 2023-06-09 at 12 44 39

codefromthecrypt:
Screenshot 2023-06-09 at 11 53 59

maybe re-kick the action?

sanposhiho:
Well, seems we need to have the label anyway then to let the bot merge the PR. 😓

anuraaga:
@sanposhiho I'm not familiar with the bot but my guess would be you need to add `/label tide/merge-method-squash` to this PR itself to switch it to non-merge mode

kerthcet:
My concern is independent commits will also be squashed. Prefer to label that manually.

codefromthecrypt:
> My concern is independent commits will also be squashed. Prefer to label that manually.

if we do well one PR is small enough that independent commits aren't needed. how about it?

sanposhiho:
Have a chat with @codefromthecrypt, let's keep as it is now. 
I'd close this.

/close

PR #34, opened 2023-06-09
Reuse call stack when invoking filter

anuraaga:
#### What type of PR is this?

/label tide/merge-method-squash

Preallocates a call stack for wasm invocation to reduce allocations during the filter hot path. This improves performance of function invocation a little bit.

This project may not be at a stage for such small optimizations so no need to proceed if so - I saw some perf work happening so figured it may help to send it

```release-note
NONE
```

```benchstat
goos: darwin
goarch: arm64
                                            │   old.txt   │              new.txt              │
PluginFilter/noop/params:_small-10            201.1n ± 3%   190.6n ± 3%  -5.17% (p=0.002 n=6)
PluginFilter/noop/params:_real-10             200.9n ± 3%   189.5n ± 2%  -5.67% (p=0.002 n=6)
PluginFilter/filter-simple/params:_small-10   8.588µ ± 3%   8.249µ ± 4%  -3.95% (p=0.041 n=6)
PluginFilter/filter-simple/params:_real-10    58.85µ ± 3%   59.30µ ± 2%       ~ (p=0.589 n=6)
geomean                                       2.126µ        2.050µ       -3.54%
```

### DIFF
--- a/scheduler/plugin/guest.go
+++ b/scheduler/plugin/guest.go
@@ -33,9 +33,10 @@ const (
 )
 
 type guest struct {
-	guest    wazeroapi.Module
-	out      *bytes.Buffer
-	filterFn wazeroapi.Function
+	guest     wazeroapi.Module
+	out       *bytes.Buffer
+	filterFn  wazeroapi.Function
+	callStack []uint64
 }
 
 func compileGuest(ctx context.Context, runtime wazero.Runtime, guestBin []byte) (guest wazero.CompiledModule, err error) {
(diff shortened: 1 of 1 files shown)


codefromthecrypt (approved):
great idea. the guest is only used sequentially, so any functions we need we can pre-size the buffer

sanposhiho (approved):
/approve
/lgtm

codefromthecrypt:
thanks for lighting turn-around as this drifts my local stuff and I was a bit anxious when it would land. All good!

PR #35, opened 2023-06-09
implements ScorePlugin

codefromthecrypt:
#### What type of PR is this?

/kind feature

This implements ScorePlugin. Notably, this reduces the size of the score returnable by wasm to int32 from int64 for reasons elaborated in the RATIONALE.md.

Fixes #

Thanks a lot to @anuraaga and @ncruces for helping with the wasm aspects of the rationale.

Thanks a lot to @kerthcet and @sanposhiho for helping with the value space aspects of the score plugin.

Only notable behavior is reducing the value range of score produced by wasm plugins to max int32.

```
goos: darwin
goarch: arm64
                                            │   old.txt   │               new.txt               │
                                            │   sec/op    │   sec/op     vs base                │
PluginFilter/noop/params:_small-12            179.5n ± 0%   181.0n ± 0%  +0.84% (p=0.002 n=6)
PluginFilter/noop/params:_real-12             178.8n ± 1%   182.3n ± 1%  +1.96% (p=0.004 n=6)
PluginFilter/filter-simple/params:_small-12   6.750µ ± 2%
PluginFilter/filter-simple/params:_real-12    50.62µ ± 2%
PluginFilter/noop-wat/params:_small-12                      120.1n ± 0%
PluginFilter/noop-wat/params:_real-12                       119.8n ± 0%
PluginFilter/wasm/params:_small-12                          6.671µ ± 1%
PluginFilter/wasm/params:_real-12                           50.26µ ± 2%
PluginScore/noop-wat/params:_small-12                       120.5n ± 2%
PluginScore/noop-wat/params:_real-12                        122.0n ± 1%
PluginScore/noop/params:_small-12                           244.5n ± 4%
PluginScore/noop/params:_real-12                            206.5n ± 1%
PluginScore/wasm/params:_small-12                           5.082µ ± 5%
PluginScore/wasm/params:_real-12                            49.80µ ± 1%
geomean                                       1.820µ        746.4n       +1.39%               ¹
¹ benchmark set differs from baseline; geomeans may not be comparable
```


RATIONALE.md
```diff
@@ -0,0 +1,61 @@
...
+
+We currently use 32-bits to return the status as an i32. It is still easy to
+return a second 32-bit value because two can be packed into a single i64.
+
+When the combined results require more than 64 bits, we have to share memory to
+return more results, as out pointer parameters. This is complicated when the
+function is called host->guest, because the guest controls memory. For the host
+to pass an out-pointer, it needs a region of memory unused by the guest.
+
+Most often, this is reserved with exported functions. For example, TinyGo
```
codefromthecrypt, line 26: inspired by @ncruces approach with sqlite (reserve memory for whole module use) we could also make a single init function which the host passes how much scratch space. the guest keeps an active pointer to this the whole life of the module. Free isn't needed because it is reserved the entire time. This would be toolchain independent and work for many fixed width things like score, but not for things like name which are unbounded.

codefromthecrypt: I think for score alone, this is overkill, but anyway food for thought in case the topic comes up again.

guest/api/types.go
```diff
@@ -31,23 +31,24 @@ var _ FilterPlugin = FilterFunc(nil)
...
 
-// Filter returns f(a).
+// Filter returns f(pod, nodeInfo).
 func (f FilterFunc) Filter(pod Pod, nodeInfo NodeInfo) *Status {
 	return f(pod, nodeInfo)
 }
 
-// ScorePlugin is a WebAssembly implementation of framework.ScorePlugin.
+// ScorePlugin is a WebAssembly implementation of framework.ScorePlugin except
+// the score is uint32, not int64.
```
anuraaga, line 40: Looking at the PR I wonder if `int32` may be better that `uint32`. I have seen scoring situations before where default is zero, some things are higher priority than default and scores within positive range, others are lower priority than default but still scored within negative range. Whether this is something actually done in k8s I don't know but I guess matching the standard type semantically can prevent confusion, while restricting the precision because 64-bits is overkill and anyways most math in wasm will be 32-bits still shouldn't be too confusing / limiting.

Anyways mostly an observation as food for thought, definitely not an expert on it :)

guest/score/score_exports.go
```diff
@@ -0,0 +1,49 @@
...
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/imports"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/types"
+)
+
+// score is only exported to the host.
+//
+//export score
+func score() uint64 {
+	// Zero on unconfigured score
```
anuraaga, line 31: This is what made me think of the other comment

scheduler/plugin/guest.go
```diff
@@ -67,26 +68,47 @@ func (pl *wasmPlugin) newGuest(ctx context.Context) (*guest, error) {
 		out.Reset()
 	}
 
+	// Allocate a call stack sized to max of params / return values of any
```
anuraaga, line 71: Thanks, didn't realize more functions were coming so soon 😁

Because of that I didn't implement an idea of computing the size in `detectExports`, just an idea probably for another PR anyways

scheduler/plugin/host.go
```diff
@@ -26,79 +26,101 @@ import (
...
 	bufLimit := bufLimit(stack[1])
 
-	pod := filterParamsFromContext(ctx).pod
+	pod := paramsFromContext(ctx).pod
 	stack[0] = uint64(marshalIfUnderLimit(mod.Memory(), pod, buf, bufLimit))
 }
+
+// k8sSchedulerStatusReasonFn is a function used by the wasm guest to set the
+// framework.Status reason.
+func k8sSchedulerStatusReasonFn(ctx context.Context, mod wazeroapi.Module, stack []uint64) {
```
anuraaga, line 114: So we do actually have a host function ingesting a status. I suppose the reason this doesn't also just go ahead and take a int64 status to match native is we expect most non-zero statuses to not have reason set by the user (zero status can be just not calling the function at all) so the performance advantage of avoiding a host call is large.

If so it _might_ be something to add to the rationale perhaps others would have a similar question.

anuraaga (commented):
Just some optional comments, lgtm

guest/api/types.go
kerthcet: In theory, yes, we have no idea how people design their plugins. Although after normalizing, they will be in the range of [0, 100]

kerthcet (commented):
/approve

codefromthecrypt:
> In theory, yes, we have no idea how people design their plugins. Although after normalizing, they will be in the range of [0, 100]

ok I will put the note in that normalizing means [0, 100] so basically regardless of number above 100 it normalizes to 100.

guest/api/types.go
codefromthecrypt: No strong opinions. I made it uint32 randomly, I admit it ;) I'll move to int32, and also add a test that negative is visible.

scheduler/plugin/guest.go
codefromthecrypt: yep, the route I'm trying for is (and seeing how big even one change is!) to contain features knowing they are incomplete or small debt is accumulating. Like the two we've implemented are missing a parameter, then solve for that specific parameter as a PR. In that line, once we have a different size param/return sum we can solve it then.

scheduler/plugin/host.go
codefromthecrypt: I'll add rationale for the reduced width of status, which is that this is for custom plugins, so the custom range you can use over the built-in status codes is only 2 million ;)

codefromthecrypt:
So, next steps for me are:
* make sure the score and status results are signed, even if narrowed to 32-bits. This increases parity even, even if negative values aren't semantically defined for either.
* consolidate confusingly large types mapped to also too large types, but more convenient ones, into a section. It seems a trend to accidentally (int64) or by default (int) use really large types for values never intended to be large).    
  * This allows us to refer to that section in places handling type differences.
  * Add to this section that especially in score it is unnecessary as the usable range is [0, 100]
  * Add to this section that the user can only define 2 million custom positive status codes now ;)
* restructure the notes currently specifically about score into a general note about when returning two values.
* fill some tests.

Will push a commit to include this!

codefromthecrypt:
ok spent another hour plus on docs. I think that's way over budget for me and unparalleled in practice.

Next is test.

codefromthecrypt:
ok addressed all the TODO I'm aware of, PTAL!

scheduler/plugin/guest.go
```diff
@@ -67,24 +68,45 @@ func (pl *wasmPlugin) newGuest(ctx context.Context) (*guest, error) {
...
+// score calls guestExportScore.
+func (g *guest) score(ctx context.Context) (int64, *framework.Status) {
+	defer g.out.Reset()
+	callStack := g.callStack
+
+	if err := g.scoreFn.CallWithStack(ctx, callStack); err != nil {
+		return 0, framework.AsStatus(decorateError(g.out, "score", err))
+	}
+
+	score := int32(callStack[0] >> 32)
```
codefromthecrypt, line 106: added tests to make sure this part wasn't goofed (e.g. cast to uint32 was here until we switched to signed)

codefromthecrypt:
once last feedback is in, I will squash so that the commit history isn't junked up

anuraaga (approved):

guest/api/types.go
sanposhiho: Yes, even in in-tree plugins, I remember some of them return negative scores in Score and then make it in [0, 100] in NormalizeScore(). So, int32 LGTM

sanposhiho (commented):
Thanks! LGTM, leave one nit fix and one question. 
Please go squashing.

guest/score/score_exports.go
```diff
@@ -0,0 +1,49 @@
...
+	if Plugin == nil {
```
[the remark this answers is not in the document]
codefromthecrypt: actually a little different. So this is the guest side. Anyone importing this package will export the function, so the host will think it should be run. I think this could be changed to a panic because if it is still nil, it means someone imported the package, but didn't assign a function to it.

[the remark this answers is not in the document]
codefromthecrypt: ok switched it to panic because not setting this is a bug.

examples/score-simple/go.mod
```diff
@@ -1,4 +1,4 @@
-module sigs.k8s.io/kube-scheduler-wasm-extension/examples/noop
```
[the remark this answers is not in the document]
codefromthecrypt: I moved the go.mod up one to remove this sort of drift

codefromthecrypt:
thanks for the looks folks!

guest/filter/filter_exports.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// filter is only exported to the host.
+//
+//export filter
+func filter() uint32 { //nolint
+	if Plugin == nil {
+		// If we got here, someone imported the package, but forgot to set the
+		// filter. Panic with what's wrong.
+		panic("filter imported, but filter.Plugin nil")
```
codefromthecrypt, line 33: this is new. before we silently returned success, but in reality there's no reason to import this package and not set the plugin.

sanposhiho: Thanks!

sanposhiho (approved):
/lgtm
/unhold

PR #37, opened 2023-06-14
fix(host): remove lock from Score

sanposhiho:
#### What type of PR is this?

/kind cleanup

The scheduling framework doesn't call Score() of the same plugin in parallel. So, we don't need to have `pl.pool.assignedToSchedulingPodLock.Lock()`, it's the lock to protect the guest from using at the same time.

ref:  https://github.com/kubernetes/kubernetes/blob/be14b026e33b682c52750741a2256373e480b7d7/pkg/scheduler/framework/runtime/framework.go#L1037

Fixes #

```release-note
NONE
```

```benchstat
(trivial)
```


codefromthecrypt (approved):
thanks!

PR #39, opened 2023-06-15
Switches from UID to pointer for identifying cycle IDs

codefromthecrypt:
#### What type of PR is this?

/kind bug

Before, we used the UID to identify which cycle is in progress. However, when a scheduling cycle fails, the same UID can happen the next time.

This uses the pointer instead as this is different per run, and easier to compare in wasm. We use the lower bits as that's likely to not conflict, easier to use in wasm, and also can't identify fully the real host memory address.

Finally, this resets the pointer between benchmark runs. This makes sure a run isn't accidentally faster. Before, it was faster due to colliding on the same UID. By resetting the pointer, we avoid re-creating that issue.

NONE

I noticed curious results in the prior PR. The current results look a lot more realistic

NONE

```
goos: darwin
goarch: arm64
BenchmarkPluginFilter
BenchmarkPluginFilter/noop-wat
BenchmarkPluginFilter/noop-wat/params:_small
BenchmarkPluginFilter/noop-wat/params:_small-12         	 4514036	       271.1 ns/op
BenchmarkPluginFilter/noop-wat/params:_real
BenchmarkPluginFilter/noop-wat/params:_real-12          	 4434201	       263.2 ns/op
BenchmarkPluginFilter/noop
BenchmarkPluginFilter/noop/params:_small
BenchmarkPluginFilter/noop/params:_small-12             	 3605126	       335.3 ns/op
BenchmarkPluginFilter/noop/params:_real
BenchmarkPluginFilter/noop/params:_real-12              	 3170498	       339.0 ns/op
BenchmarkPluginFilter/test
BenchmarkPluginFilter/test/params:_small
BenchmarkPluginFilter/test/params:_small-12             	  154036	      7219 ns/op
BenchmarkPluginFilter/test/params:_real
BenchmarkPluginFilter/test/params:_real-12              	    9352	    121479 ns/op
BenchmarkPluginScore
BenchmarkPluginScore/noop-wat
BenchmarkPluginScore/noop-wat/params:_small
BenchmarkPluginScore/noop-wat/params:_small-12          	 4664058	       257.6 ns/op
BenchmarkPluginScore/noop-wat/params:_real
BenchmarkPluginScore/noop-wat/params:_real-12           	 4602486	       261.2 ns/op
BenchmarkPluginScore/noop
BenchmarkPluginScore/noop/params:_small
BenchmarkPluginScore/noop/params:_small-12              	 3169807	       377.1 ns/op
BenchmarkPluginScore/noop/params:_real
BenchmarkPluginScore/noop/params:_real-12               	 3494535	       344.6 ns/op
BenchmarkPluginScore/test
BenchmarkPluginScore/test/params:_small
BenchmarkPluginScore/test/params:_small-12              	  283779	      4048 ns/op
BenchmarkPluginScore/test/params:_real
BenchmarkPluginScore/test/params:_real-12               	   22026	     54050 ns/op
BenchmarkPluginFilterAndScore
BenchmarkPluginFilterAndScore/noop-wat
BenchmarkPluginFilterAndScore/noop-wat/params:_small
BenchmarkPluginFilterAndScore/noop-wat/params:_small-12 	 3251036	       367.1 ns/op
BenchmarkPluginFilterAndScore/noop-wat/params:_real
BenchmarkPluginFilterAndScore/noop-wat/params:_real-12  	 3233397	       369.3 ns/op
BenchmarkPluginFilterAndScore/noop
BenchmarkPluginFilterAndScore/noop/params:_small
BenchmarkPluginFilterAndScore/noop/params:_small-12     	 2127667	       565.0 ns/op
BenchmarkPluginFilterAndScore/noop/params:_real
BenchmarkPluginFilterAndScore/noop/params:_real-12      	 2246898	       531.6 ns/op
BenchmarkPluginFilterAndScore/test
BenchmarkPluginFilterAndScore/test/params:_small
BenchmarkPluginFilterAndScore/test/params:_small-12     	  105128	     10904 ns/op
BenchmarkPluginFilterAndScore/test/params:_real
BenchmarkPluginFilterAndScore/test/params:_real-12      	    5520	    211922 ns/op
```

### DIFF
--- a/scheduler/plugin/plugin.go
+++ b/scheduler/plugin/plugin.go
@@ -165,31 +165,29 @@ func (pl *wasmPlugin) PreFilterExtensions() framework.PreFilterExtensions {
 
 // PreFilter implements the same method as documented on
 // framework.PreFilterPlugin.
-func (pl *wasmPlugin) PreFilter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod) (*framework.PreFilterResult, *framework.Status) {
-	_, err := pl.pool.getForScheduling(ctx, cycleID(pod))
-	if err != nil {
-		return nil, framework.AsStatus(err)
+func (pl *wasmPlugin) PreFilter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod) (result *framework.PreFilterResult, status *framework.Status) {
+	if err := pl.pool.doWithSchedulingGuest(ctx, cycleID(pod), func(g *guest) {
+		// TODO: partially implemented for testing
+	}); err != nil {
+		status = framework.AsStatus(err)
 	}
-
-	// TODO: partially implemented for testing
-
-	return nil, nil
+	return
 }
 
 var _ framework.FilterPlugin = (*wasmPlugin)(nil)
 
 // Filter implements the same method as documented on framework.FilterPlugin.
-func (pl *wasmPlugin) Filter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
-	g, err := pl.pool.getForScheduling(ctx, cycleID(pod))
-	if err != nil {
-		return framework.AsStatus(err)
-	}
-
+func (pl *wasmPlugin) Filter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) (status *framework.Status) {
 	// Add the params to the go context so that the corresponding host function
 	// can look them up.
 	params := &params{pod: pod, nodeInfo: nodeInfo}
 	ctx = context.WithValue(ctx, paramsKey{}, params)
-	return g.filter(ctx)
+	if err := pl.pool.doWithSchedulingGuest(ctx, cycleID(pod), func(g *guest) {
+		status = g.filter(ctx)
+	}); err != nil {
+		status = framework.AsStatus(err)
+	}
+	return
 }
 
 var _ framework.PostFilterPlugin = (*wasmPlugin)(nil)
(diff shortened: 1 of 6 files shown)


scheduler/plugin/pool.go
```diff
@@ -19,124 +19,131 @@ package wasm
...
-
-	// assignedToSchedulingPodLock is a lock to protect the access to the assigned instance.
-	// The scheduler may call Filter(), AddPod(), RemovePod() concurrently, and we need to take a lock.
-	// But, other methods of the plugin should be invoked for the same Pod serially.
-	assignedToSchedulingPodLock sync.Mutex
-
-	// assignedToBindingPod has the guest instances for Pods in binding cycle.
-	assignedToBindingPod map[types.UID]guest
-}
+	mux sync.RWMutex
```
codefromthecrypt, line 32: so this iteration uses a simple lock for now. This is under the assumption that access to one plugin isn't concurrent enough to warrant the additional complexity of read and write locks, where often you need a write one. We can later optimize this, but for now, simple IMHO is better.

scheduler/plugin/plugin.go
```diff
@@ -191,10 +180,7 @@ var _ framework.FilterPlugin = (*wasmPlugin)(nil)
 
 // Filter implements the same method as documented on framework.FilterPlugin.
 func (pl *wasmPlugin) Filter(ctx context.Context, _ *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
-	pl.pool.assignedToSchedulingPodLock.Lock()
-	defer pl.pool.assignedToSchedulingPodLock.Unlock()
-
-	g, err := pl.pool.getOrCreateGuest(ctx, pod.GetUID())
+	g, err := pl.pool.getForScheduling(ctx, cycleID(pod))
```
sanposhiho, line 183: `assignedToSchedulingPodLock` was the lock to protect guest to be used in parallel. (because Filter can be called in parallel during preemption)
And here, because you remove `assignedToSchedulingPodLock`, the guest getting here can be used in parallel during preemption, right? maybe we should keep `assignedToSchedulingPodLock`?

sanposhiho (commented):
I haven't yet gone through all but left one comment for now.

scheduler/plugin/plugin.go
codefromthecrypt: I see. indeed the pool lock won't prevent overlapping calls to use the guest.

So I think I need to revise the docs because scheduling is also not one at a time.

codefromthecrypt: I think state problem is too abstract for me to have brought up, yet because we don't implement it in this plugin, yet.

I will fix the overlapping problem on this PR and bear in mind the details about state later!

codefromthecrypt: I changed the impl to serialize access during scheduling, and put some notes about my understanding of it. Thanks for all the help and feel free to correct more

sanposhiho (approved):
/lgtm
/approve

codefromthecrypt:
Thanks, @sanposhiho I learned a lot too, and any tricks to identify pre-emption sound really useful. I think we'll end up revisiting the design a couple times at least, especially when we get pre-emption tests in!

PR #58, opened 2023-07-17
allow the scheduler configuration to have multi wasm plugins

### LINKED ISSUE #23: make it possible to load multiple wasm plugin
**What would you like to be added**:

Currently, the host scheduler has one wasm plugin, and it can load one wasm guest.
Meaning, basically people can have only one wasm.
It'd be great if users can use multiple wasm guest.

**Why is this needed**:

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

sanposhiho:
#### What type of PR is this?

/kind feature

This PR changes the entry point to read the scheduler configuration before it initializes the scheduler (app.NewSchedulerCommand) and allows us to have multiple wasm plugins.

```yaml
apiVersion: kubescheduler.config.k8s.io/v1
kind: KubeSchedulerConfiguration
profiles:
  - plugins:
      multipoint:
        enabled:
           - name: wasmplugin1
           - name: wasmplugin2
     pluginConfig:
       - name: wasmplugin1
          args:
            guestPath: "/path/to/wasm-plugin1"
       - name: wasmplugin2
          args:
            guestPath: "/path/to/wasm-plugin2"
```

Fixes #23

```release-note
The scheduler can have multiple wasm plugins.
```

```benchstat
N/A (no benchmark related area in this change.)
```

### DIFF
--- a/scheduler/cmd/scheduler/config.go
+++ b/scheduler/cmd/scheduler/config.go
@@ -1,0 +1,99 @@
+/*
+   Copyright 2023 The Kubernetes Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	_ "k8s.io/component-base/logs/json/register" // for JSON log format registration
+	_ "k8s.io/component-base/metrics/prometheus/clientgo"
+	_ "k8s.io/component-base/metrics/prometheus/version" // for version metric registration
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config/scheme"
+	frameworkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+
+	wasm "sigs.k8s.io/kube-scheduler-wasm-extension/scheduler/plugin"
+)
+
+// getWasmPluginsFromConfig parses the scheduler configuration specified with --config option,
+// and return the wasm plugins enabled by the user.
+func getWasmPluginsFromConfig() ([]string, error) {
+	// In the scheduler, the path to the scheduler configuration is specified with --config option.
+	configFile := flag.String("config", "", "")
+	flag.Parse()
+
+	if configFile == nil {
+		// Users don't have the own configuration. do nothing.
+		return nil, nil
+	}
+
+	cfg, err := loadConfigFromFile(*configFile)
+	if err != nil {
+		return nil, err
+	}
+
+	return getWasmPluginNames(cfg), nil
+}
+
+// getWasmPluginNames returns the wasm plugin names enabled by the user.
+// It assumes that the wasm plugin is specified as the multi-point plugin.
+func getWasmPluginNames(cc *config.KubeSchedulerConfiguration) []string {
+	names := []string{}
+	for _, profile := range cc.Profiles {
+		wasmplugins := sets.New[string]()
+		// look for the wasm plugin in the plugin config.
+		for _, config := range profile.PluginConfig {
+			if err := frameworkruntime.DecodeInto(config.Args, &wasm.PluginConfig{}); err != nil {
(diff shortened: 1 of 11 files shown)


scheduler/cmd/scheduler/main.go
```diff
@@ -17,21 +17,90 @@
...
+}
+
+// getWasmPluginNames returns the wasm plugin names enabled by the user.
+// It assumes that the wasm plugin is specified as the multi-point plugin,
+// and the wasm plugin name starts with wasm.PluginNamePrefix.
+func getWasmPluginNames(cc *config.KubeSchedulerConfiguration) []string {
+	wasmplugins := []string{}
+	for _, profile := range cc.Profiles {
+		// assuming that the wasm plugin is specified as a multi-point plugin.
+		for _, plugin := range profile.Plugins.MultiPoint.Enabled {
```
codefromthecrypt, line 81: Is this enabled by default? It is cool that there is a concept of having multiple plugins, but I wonder why we would do this which requires a name prefix?

If we don't use "multipoint" we could use another nested property, allow the name to be plain, and not accidentally turned off. For example, look for guestPath instead and don't try to fit it into the MultiPoint thing.

Then, everything in the plugin is handled normally, it seems. What am I missing?

sanposhiho:
After spending good days in NZ, my capacity is coming back here. 
Will address the comment from Adrian and open it then.

scheduler/cmd/scheduler/main.go
sanposhiho: Removed a name prefix.

sanposhiho:
OK, it's ready.
@kerthcet Can you please take a look when have time.

scheduler/cmd/scheduler/config_test.go
```diff
@@ -0,0 +1,128 @@
...
+									Args: &config.DefaultPreemptionArgs{},
+								},
+								{
+									Name: "InterPodAffinity",
+									Args: &config.InterPodAffinityArgs{},
+								},
+								{
+									Name: "wasm1",
+									Args: &runtime.Unknown{
+										// TODO: need to make the wasm config implements runtime.Object.
```
Gekko0114, line 99: What is needed here? Is this PR not sufficient to work yet?

scheduler/plugin/plugin.go
```diff
@@ -136,7 +131,7 @@ var _ framework.Plugin = (*wasmPlugin)(nil)
 // Name implements the same method as documented on framework.Plugin.
 // See /RATIONALE.md for impact
```
Gekko0114, line 132: Maybe need to update RATIONALE.md?

Gekko0114:
Minor comment


> apiVersion: kubescheduler.config.k8s.io/v1
> kind: KubeSchedulerConfiguration
> profiles:
>   - plugins:
>       multipoint:
>         enabled:
>            - name: wasmplugin1
>            - name: wasmplugin2
>      pluginConfig:
>        - name: wasmplugin1
>           args:
>             guestPath: "/path/to/wasm-plugin1"
>        - name: wasmplugin2
>           args:
>             guestPath: "/path/to/wasm-plugin2"

guestPath should have extension .wasm?

scheduler/cmd/scheduler/config.go
```diff
@@ -0,0 +1,99 @@
...
+				continue
+			}
+
+			wasmplugins.Insert(config.Name)
+		}
+
+		// look for the wasm plugin in the enabled plugins.
+		// (assuming that the wasm plugin is specified as a multi-point plugin.)
+		for _, plugin := range profile.Plugins.MultiPoint.Enabled {
+			if wasmplugins.Has(plugin.Name) {
```
Gekko0114, line 74: If the plugin specified by plugins.multipoint.enabled does not exist in wasmplugins, it might be good to log something.

Gekko0114:
Minor comment: 
It might be better to have a test case that verifies correct execution by passing multiple wasm files.

Gekko0114:
Overall, it looks very good :) I have left several minor comments

scheduler/cmd/scheduler/config_test.go
sanposhiho: This PR works fine without getting this TODO done.
But, if we make the wasm scheduler config implements runtime.Object, we don't need to use runtime.Unknown, instead we can do something like InterPodAffinityArgs above which would look cleaner:

```
{
	Name: "InterPodAffinity",
	Args: &config.InterPodAffinityArgs{},
},
```

scheduler/plugin/plugin.go
sanposhiho: Good point, will give update there

scheduler/cmd/scheduler/config.go
sanposhiho: I wouldn't rather do that.

sanposhiho:
@Gekko0114 fixed or replied. PTAL again.

Gekko0114:
LGTM!

PR #59, opened 2023-07-18
Implement PreScore and change NodeNumberPlugin to use it

codefromthecrypt:
#### What type of PR is this?

/kind feature

This implement PreScore and changes NodeNumberPlugin to use it.

N/A

Almost done: I need to do a review pass tomorrow.

NONE

Performance is slightly slower because we always call prefilter. So, when we move logic to prescore, there's an additional host call roundtrip.

```benchstat
goos: darwin
goarch: arm64
                          │ before.txt  │              now.txt              │
Example_NodeNumber/New-12   63.43m ± 4%   64.32m ± 1%       ~ (p=0.394 n=6)
Example_NodeNumber/Run-12   51.34µ ± 2%   51.88µ ± 2%       ~ (p=0.058 n=6)
geomean                     1.805m        1.827m       +1.23%
```


guest/api/proto/proto.go
```diff
@@ -22,6 +22,11 @@ import (
 	meta "sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/meta"
 )
 
+type Nodes interface {
```
codefromthecrypt, line 25: thinking about exposing the slice like this as there's no lazy slice in Go. Will try it out ergonomically once finished other stuff.

examples/nodenumber/plugin/plugin_test.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"testing"
+
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api/proto"
+	protoapi "sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/api"
+	meta "sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/meta"
+)
+
+func Test_NodeNumber(t *testing.T) {
```
codefromthecrypt, line 28: these tests run both in Go and TinyGo, putting the business logic in a non-main package helped avoid constraints around wasm imports and nottinygc

codefromthecrypt:
going with this first to keep PreScore change simple https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/60

codefromthecrypt:
ok once this is in, I'll implement configuration to make our NodeNumberPlugin more realistic.

scheduler/plugin/plugin.go
```diff
@@ -275,13 +275,28 @@ func (pl *wasmPlugin) PostFilter(ctx context.Context, state *framework.CycleStat
...
 	// We implement PreScorePlugin with ScorePlugin, even when the guest doesn't.
 	if pl.guestInterfaces&iPreScorePlugin == 0 {
 		return nil // unimplemented
 	}
 
-	panic("TODO: scheduling: PreScore")
+	// Add the stack to the go context so that the corresponding host function
+	// can look them up.
+
+	// Use v1.NodeList to encode the nodes, as it is easier for both sides.
```
codefromthecrypt, line 287: I will make one change to lazy encode this, only when asked by the guest. otherwise this PR is stable

codefromthecrypt:
What's needed after this is a config ABI similar to http-wasm. I'll add links on that for someone to follow-up on:

https://http-wasm.io/http-handler-abi/#get_config
https://github.com/http-wasm/http-wasm-guest-tinygo/blob/27f49b76b63759ad5694e32545dba595cd5fe9db/handler/host.go#L22-L25
https://github.com/http-wasm/http-wasm-guest-tinygo/blob/27f49b76b63759ad5694e32545dba595cd5fe9db/handler/internal/imports/imports.go#L10-L11
https://github.com/http-wasm/http-wasm-host-go/blob/990427a52ee421643c21142d462304c91415932b/handler/middleware.go#L251-L258
https://github.com/http-wasm/http-wasm-host-go/blob/990427a52ee421643c21142d462304c91415932b/handler/options.go#L25-L30

This would be used to complete the NodeNumberPlugin and would work if the config can be read inside the wasm guest, assigned by the host.

sanposhiho (approved):
Perfection. Thanks.

/lgtm
/approve

PR #61, opened 2023-07-21
Run codec benchmarks in wasm

codefromthecrypt:
#### What type of PR is this?

/kind cleanup

This changes the guest codec benchmarks to actually run in wasm. By using TinyGo with wazero, we can see performance of unmarshal realistically. Before, the benchmarks were run in normal go, even if they used the same library as the guest used.

When you run `make bench`, tinygo compiles the benchmarks to wasm, and then runs that wasm with wazero. So, the console output is coming from wasm functions compiled from Go benchmarks.

Our makefile ensures that the wazero version is the same as what the scheduler plugin uses (via go.mod). Since the wazero command configures the module to use real clocks, the performance report should be accurate.

NONE

This doesn't solve for real data yet because I can't get the yaml to proto converter to work in TinyGo. We may want to check in serialized protos as they aren't that big.

NONE

```bash
$ make bench-guest
BenchmarkUnmarshalVT/node:_small        	 1588876	       761.6 ns/op
BenchmarkUnmarshalVT/node:_small        	 1553385	       756.4 ns/op
BenchmarkUnmarshalVT/node:_small        	 1608015	       779.7 ns/op
BenchmarkUnmarshalVT/node:_small        	 1602940	       750.1 ns/op
BenchmarkUnmarshalVT/node:_small        	 1573089	       774.9 ns/op
BenchmarkUnmarshalVT/node:_small        	 1599826	       750.2 ns/op
BenchmarkUnmarshalVT/pod:_small         	  475982	      2544 ns/op
BenchmarkUnmarshalVT/pod:_small         	  476852	      2552 ns/op
BenchmarkUnmarshalVT/pod:_small         	  478239	      2610 ns/op
BenchmarkUnmarshalVT/pod:_small         	  438692	      2533 ns/op
BenchmarkUnmarshalVT/pod:_small         	  480537	      2503 ns/op
BenchmarkUnmarshalVT/pod:_small         	  484261	      2521 ns/op
PASS
ok  	sigs.k8s.io/kube-scheduler-wasm-extension/internal/e2e/guest	20.972s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,17 @@ build-tinygo: examples/nodenumber/main.wasm guest/testdata/all/main.wasm guest/t
 	@(cd $(@D); tinygo build -o main-debug.wasm -gc=custom -tags=custommalloc -scheduler=none -target=wasi .)
 
 # Testing the guest code means running it with TinyGo, which internally
-# compiles the unit tests to a wasm binary, then runs it in a WASI runtime.
+# compiles the unit tests to a wasm binary, then runs it with wazero.
 .PHONY: test-guest
 test-guest: guest/.tinygo-target.json
 	@(cd guest; tinygo test -v -target .tinygo-target.json ./...)
 	@(cd examples; tinygo test -v -target ../guest/.tinygo-target.json ./nodenumber/plugin/...)
+
+# Benchmarking the guest code means running it with TinyGo, which internally
+# compiles the benchmarks to a wasm binary, then runs it with wazero.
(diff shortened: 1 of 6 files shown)


codefromthecrypt:
cc @anuraaga pretty cool nottinygc runs inside the benchmark runner. So, we can get a realistic performance view (since our main code use nottinygc running bench without it wouldn't be relevant)

sanposhiho (approved):
/lgtm
/approve

Thanks

PR #66, opened 2023-07-27
guest: inlines host.GetConfig to config.Get

codefromthecrypt:
#### What type of PR is this?

/kind cleanup

This inlines `host.GetConfig()` to `config.Get()` after noticing logging will be in another package. This is simpler than describing a new term "host" which is only used to get configuration.

NONE

This also does refactoring to sketch out how to test logic that requires wasm imports.

NONE

```benchstat
goos: darwin
goarch: arm64
                          │ before.txt  │              now.txt              │
Example_NodeNumber/New-12   73.68m ± 2%   75.34m ± 2%       ~ (p=0.093 n=6)
Example_NodeNumber/Run-12   46.12µ ± 4%   46.82µ ± 1%       ~ (p=0.394 n=6)
geomean                     1.843m        1.878m       +1.88%
```


guest/internal/mem/mem.go
```diff
@@ -68,27 +68,6 @@ func Update(
 	return updater(readBuf)
 }
 
-func GetBytes(fn func(ptr uint32, limit BufLimit) (len uint32)) []byte {
```
codefromthecrypt, line 71: I removed this because it is difficult to test.

guest/config/config.go
```diff
@@ -0,0 +1,58 @@
...
+package config
+
+import (
+	"unsafe"
+
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/mem"
+)
+
+var (
+	readConfig func() string = func() string {
```
codefromthecrypt, line 27: ok I reworked this to add unit tests. This is the first time it is possible to test the logic around host imports.

codefromthecrypt:
oops one more thing to settle

sanposhiho (approved):
👍 

/lgtm
/approve

PR #67, opened 2023-07-28
Add klog.Info/Error support

codefromthecrypt:
#### What type of PR is this?

/kind feature

This begins klog support with the simplest api used in plugins: `klog.Info`, `klog.InfoS`, `klog.Error` and `klog.ErrorS`

NONE

* Note: "severity" is internal in klog, and we don't need to expose it in the API, just use the same constants.
* wasm cannot pass objects to the host. This is why we have to stringify in the guest first. This is most obvious in `klog.InfoS`, where the KVs are encoded NUL-terminated to the host.
* utilities like `klog.KObj` used reflection too much which caused things like crashing the compiler or poor performance. This ports the intent more simply.

NONE

Performance of Simple is wild because it uses the default GC and the plugin is heavy due to protos. Advanced uses [nottinygc](https://github.com/wasilibs/nottinygc) and also separates out code that uses imports so that they can be more easily tested. These notes are in the [READMEs](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/examples/advanced/README.md).

The main thing here is that performance of Advanced/Run is the same when logging is disabled, and it drops 13% (7us) per Run when enabled. This shows the deferred stringification works, but also shows logging has a measurable cost even when only a couple times.

```benchstat
goos: darwin
goarch: arm64
                                       │  before.txt  │              now.txt               │
                                       │    sec/op    │    sec/op     vs base              │
Example_NodeNumber/Simple/New-12         63.64m ±  3%   62.56m ±  0%       ~ (p=0.065 n=6)
Example_NodeNumber/Simple/Run-12         223.0µ ± 87%   294.2µ ± 32%       ~ (p=0.240 n=6)
Example_NodeNumber/Advanced/New-12       76.34m ±  2%   73.95m ±  1%  -3.13% (p=0.002 n=6)
Example_NodeNumber/Advanced/Run-12       46.71µ ±  3%   47.57µ ±  1%       ~ (p=0.065 n=6)
Example_NodeNumber/Simple_Log/New-12                    62.40m ±  1%
Example_NodeNumber/Simple_Log/Run-12                    176.6µ ± 29%
Example_NodeNumber/Advanced_Log/New-12                  73.83m ±  1%
Example_NodeNumber/Advanced_Log/Run-12                  54.63µ ±  1%
geomean                                  2.667m         2.706m        +6.35%
```


examples/nodenumber/plugin/plugin.go
```diff
@@ -50,25 +50,7 @@ import (
 //     a numeric match gets a results in a lower score than a match.
 type NodeNumber struct {
 	reverse bool
-}
-
-// New creates a new NodeNumber plugin with the given jsonConfig or returns an
```
codefromthecrypt, line 55: I just moved this to the bottom like upstream has

internal/e2e/scheduler/scheduler_test.go
```diff
@@ -117,6 +117,7 @@ func BenchmarkExample_NodeNumber(b *testing.B) {
 func newNodeNumberPlugin(ctx context.Context, t e2e.Testing, reverse bool) framework.Plugin {
 	plugin, err := wasm.NewFromConfig(ctx, wasm.PluginConfig{
 		GuestPath:   test.PathExampleNodeNumber,
+		LogSeverity: -1, // disabled
```
codefromthecrypt, line 120: I tested temporarily by setting this to zero (info) and it works

guest/klog/api/k8s_references.go
```diff
@@ -0,0 +1,35 @@
+package api
```
codefromthecrypt, line 1: I will add a header and also note that this is the logic from real klog except dodging some reflection that crashes the tinygo compiler.

codefromthecrypt:
going to try to add InfoS/ErrorS now. mostly an exercise in copying logic and backfilling tests

codefromthecrypt:
polished and stubbed structured logging.

next step: implement it

guest/klog/klog.go
```diff
@@ -0,0 +1,79 @@
...
+
+func logFn(severity internal.Severity, msg []byte) {
+	ptr, size := mem.BytesToPtr(msg)
+	log(severity, ptr, size)
+	runtime.KeepAlive(msg) // keep msg alive until ptr is no longer needed.
+}
+
+// Info is a convenience that calls the same method documented on api.Klog.
+//
+// Note: Code that uses can be unit tested in normal Go, but cannot be unit
```
codefromthecrypt, line 53: I added these for parity with the normal klog package, with warnings for folks who test with tinygo. This is an acknowledgement that few will, and so these make code easier for those people.

sanposhiho:
I like this approach. Please go ahead 👍

codefromthecrypt:
OK I will finish this up, but I won't have time to implement the structured log in the same PR (will leave the signatures for a later change). Will mark ready for review when ready.

codefromthecrypt:
#69 should go first

codefromthecrypt:
Added the commit from #69 to proceed further. Will try to finish structured logging in the same PR

guest/klog/imports.go
```diff
@@ -0,0 +1,27 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package klog
+
+import "sigs.k8s.io/kube-scheduler-wasm-extension/guest/klog/internal"
+
+//go:wasmimport k8s.io/klog log
+func log(severity internal.Severity, ptr, size uint32)
```
codefromthecrypt, line 24: TODO: add structured log which includes NUL-terminated kvs

```go
//go:wasmimport k8s.io/klog log
func log(severity internal.Severity, msgPtr, msgSize, kvPtr, kvSize uint32)
```

codefromthecrypt:
Ok I've gotten this far, but something's wrong with k/v on the host side. Will look into it later

```
I0802 15:18:53.788082   34631 host.go:219] NodeNumberArgs is successfully applied
I0802 15:18:53.788393   34631 host.go:264] "execute PreScore on NodeNumber plugin" pod="(MISSING)"
I0802 15:18:53.788411   34631 host.go:264] "execute Score on NodeNumber plugin" pod="(MISSING)"
```

codefromthecrypt:
So, what's left here is..

* debug why KV isn't working on the host (only gets the key not the value)
  * after debugging, in `newNodeNumberPlugin` revert to `LogSeverity: 3, // FatalLog`
* backfill tests for klog host functions (like exist for others, possibly making a tinygo or wat file in testdata)
* See if there's a way to integration test klog in e2e/scheduler (like capture stderr, though might be annoying due to timestamps)
* Add `ExampleInfoS` like we have `ExampleInfo` with use like in NodeNumberPlugin
* Add `ExampleError` and `ExampleErrorS` like we have `ExampleInfo` from some existing code in a real plugin
* once #69 is merged rebase this

evacchi:
I pushed 4 distinct commits so that you can pick any of these changes if you think they help:

1. fix for the logging issue (just a misconfigured test)
2. capture klog and add the assertions
3. add examples -- but I am not sure I did what you had in mind: the output is empty as it was for the other examples
4. I tried to draft another plugin under `all.go` for tinygo `logPlugin` which is just the same as `params` except it should also log them, but I did not figure out where to put the actual test cases; also it felt a bit redundant because of (2); but maybe you had something different in mind

feel free to drop all of them if they are not useful!

codefromthecrypt:
Thanks for all the help @evacchi I will go through and do any buttoning up once #69 merges, which might be monday due to some of us having typhoon displacement ;)

guest/testdata/all/main.go
```diff
@@ -37,11 +38,17 @@ type extensionPoints interface {
...
-		plugin = paramsPlugin{}
-	} else {
-		plugin = noopPlugin{}
+	var plugin extensionPoints = noopPlugin{}
+
+	if len(os.Args) == 2 {
+		switch os.Args[1] {
+		case "params":
+			plugin = paramsPlugin{}
+		case "log":
```
codefromthecrypt, line 47: @evacchi I like this approach I will tidy it up. thanks!

codefromthecrypt:
squashed work so far and will move this to the finish line soon. Thanks a lot @evacchi!

codefromthecrypt:
@sanposhiho I'm marking ready for review even if I have some test cases to backfill. You can see in the PR desc that logging enabled vs disabled makes a difference in performance, which means our ABI is working! Thanks to @evacchi for the collaboration on this.

guest/testdata/all/main.go
codefromthecrypt: this still needs to actually be invoked. I will TODO on this prior to merge

sanposhiho (approved):
It's a giant step. Thanks for the effort @codefromthecrypt @evacchi!
I left two tiny nits, feel free to merge after them.

/approve

codefromthecrypt:
ok ready. only other refactoring will be separate PR as I want to see if we can avoid using "wat" format too much.

The main thing here is we can see that the advanced plugin perf is within the margin of error (comparable to before), and when logging is switched on, it drops 13% (7us) per run, which makes sense as logging is expensive.

guest/testdata/all/main.go
codefromthecrypt, line 1: deleted this as it became complicated and almost all complicated logic wasn't used (because we moved to benchmarking examples as a key metric)

scheduler/plugin/host_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/tetratelabs/wazero/experimental/wazerotest"
+	"k8s.io/klog/v2"
+	k8stest "k8s.io/klog/v2/test"
+)
+
+func Test_k8sKlogLogFn(t *testing.T) {
```
codefromthecrypt, line 29: @sanposhiho these tests are new, using "wazerotest" which might be released in the next version of wazero. If you are cool `/lgtm` or otherwise lemme know. I intentionally kept them simple (vs using tables etc) for the first go.

cc @ncruces and @achille-roussel the first to use this style!

sanposhiho: Cool!

sanposhiho (approved):
/lgtm
/approve

PR #68, opened 2023-07-28
config: add support to file:// and http(s):// URIs

evacchi:
#### What type of PR is this?

/kind feature

/kind api-change

Extends the current plugin config to use instead a URI. In the case of `file://` the behavior is the same as it is currently. In the case of `http(s)://` it will fetch the URI and try to evaluate it as a wasm payload.

This PR is based on earlier work on `dapr/component-contrib`. See: https://github.com/dapr/components-contrib/pull/3005

n/a will create if required.

n/a

```release-note

Changes `PluginConfig.ConfigPath` to `PluginConfig.ConfigURL`. A local path must now be prefixed with `file://`. A local path will be of the form `file://relativepath/myfile.wasm` or `file:///absolute/path/to/myfile.wasm`.

```

n/a

### DIFF
--- a/scheduler/plugin/http.go
+++ b/scheduler/plugin/http.go
@@ -46,18 +46,19 @@ func (f *httpClient) get(ctx context.Context, u *url.URL) ([]byte, error) {
 		return nil, err
 	}
 	resp, err := f.c.Do(req.WithContext(ctx))
+	defer func() {
+		io.Copy(io.Discard, resp.Body) //nolint
+		resp.Body.Close()
+	}()
 	if err != nil {
 		return nil, err
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		io.Copy(io.Discard, resp.Body) //nolint
-		resp.Body.Close()
 		return nil, fmt.Errorf("received %v status code from %q", resp.StatusCode, u)
 	}
 
 	bytes, err := io.ReadAll(resp.Body)
-	resp.Body.Close()
 	if err != nil {
 		return nil, err
 	}
--- a/scheduler/plugin/http.go
+++ b/scheduler/plugin/http.go
@@ -1,0 +1,65 @@
+/*
+   Copyright 2023 The Kubernetes Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package wasm
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+)
+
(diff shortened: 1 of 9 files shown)


scheduler/plugin/http.go
```diff
@@ -0,0 +1,65 @@
...
+// It is possible to plug a custom http.RoundTripper to handle other concerns (e.g. retries)
+// Compression is handled transparently and automatically by http.Client.
+func newHTTPCLient(transport http.RoundTripper) *httpClient {
+	return &httpClient{
+		c: http.Client{Transport: transport},
+	}
+}
+
+// fetch returns a byte slice of the wasm module found at the given URL, or an error otherwise.
+func (f *httpClient) get(ctx context.Context, u *url.URL) ([]byte, error) {
```
sanposhiho, line 43: nit

```suggestion
// newHTTPClient is a constructor for httpFetcher.
//
// It is possible to plug a custom http.RoundTripper to handle other concerns (e.g. retries)
// Compression is handled transparently and automatically by http.Client.
func newHTTPClient(transport http.RoundTripper) *httpClient {
	return &httpClient{
		c: http.Client{Transport: transport},
	}
}

// get returns a byte slice of the wasm module found at the given URL, or an error otherwise.
func (f *httpClient) get(ctx context.Context, u *url.URL) ([]byte, error) {
```

scheduler/plugin/http.go
```diff
@@ -0,0 +1,65 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := f.c.Do(req.WithContext(ctx))
+	if err != nil {
+		return nil, err
+	}
+
```
sanposhiho, line 52: nit: Let's do `defer resp.Body.Close()` here instead of L55 and L60.

sanposhiho (approved):
Thanks, I like this. It's almost `/lgtm`, I left small nits.

/approve

scheduler/plugin/http.go
evacchi: I'll also defer `io.Copy(io.Discard, resp.Body)` as I've been told `resp.Body.Close()` may take a longer time if it has not been consumed first

codefromthecrypt:
Thanks @evacchi /lgtm

scheduler/plugin/http.go
codefromthecrypt: I would like to remove the last part in a later change for the reason added in a comment to #69. In general, we really should not do anything more than normal because it code is tech debt until proven otherwise.

```
	// golang/go#60240 recommends the following, but we are draining anyway:
	// > I would not recommend going out of your way to drain response bodies
	// > unless you've got a specific and measurable performance issue tha
	// > you're trying to resolve.to not go out of your way to drain the body.
	//
	// TODO: Consider the above advice and remove the draining, or explicitly
	// state why we are draining against advice, considering this client is
	// rarely used (only to grab the wasm during init: keep-alive in case of
	// error is even less likely to be helpful, and certainly not unit tested).
	if resp.StatusCode != http.StatusOK {
		_, _ = io.Copy(io.Discard, resp.Body)
		return nil, fmt.Errorf("received %v status code from %q", resp.StatusCode, url)
	}
```

codefromthecrypt: the specific concern about `resp.Body.Close()` taking longer or not, for example, really isn't a concern in a function whose only purpose is to read the whole body.

PR #69, opened 2023-08-01
guest: add plugin.Set for simpler example and fixes URL parsing

codefromthecrypt:
#### What type of PR is this?

/kind cleanup

This adds `plugin.Set` which makes it easier to write plugins. This also makes the example that uses it simpler by avoiding nottinygo. This is simpler because it reduces the amount of flags used, as well a flat package is easier to write. However, this can't be tested with tinygo anymore, and is slower.

Hence, this adds an "advanced" example with the former logic, which is more verbose and tricky, but faster.

NONE

This also fixes some bugs in URL parsing due to some logic not copied over from dapr. Notably, the original logic handled relative paths and better error messages. This also simplifies the logic further as we were parsing into URL type, just to convert it back to a string! Also, we started using testify by accident where that's not used anywhere else.

cc @evacchi sorry I didn't look closely at your earlier PR or would have caught this!

NONE

You can see below that the simple plugin runs >5x slower than the advanced. However, because it is still less than millisecond, some users may be ok with this.

```benchstat
goos: darwin
goarch: arm64
                                   │  before.txt  │       now.txt        │
                                   │    sec/op    │    sec/op      vs base   │
Example_NodeNumber/New-12            75.08m ±  2%
Example_NodeNumber/Run-12            47.86µ ± 10%
Example_NodeNumber/Simple/New-12                    62.23m ±   1%
Example_NodeNumber/Simple/Run-12                    188.6µ ± 109%
Example_NodeNumber/Advanced/New-12                  73.61m ±   2%
Example_NodeNumber/Advanced/Run-12                  47.03µ ±   2%
geomean                              1.895m         2.525m         ? ¹ ²
¹ benchmark set differs from baseline; geomeans may not be comparable
² ratios must be >0 to compute geomean
```


examples/nodenumber/main_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+	"testing"
+
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api/proto"
+	protoapi "sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/api"
+)
+
+func Test_NodeNumber(t *testing.T) {
```
codefromthecrypt, line 27: these tests still work in normal Go. In fact a lot of SDKs only test with Go and never TinyGo, probably because it is too complicated.

That said, I think the guest SDK itself and advanced example should stil work with `tinygo test -target=wasi` (and they do).

guest/score/score.go
```diff
@@ -58,10 +58,10 @@ var _ func() uint64 = _score
...
 //export score
 func _score() uint64 {
-	if score == nil {
-		// If we got here, someone imported the package, but forgot to set the
-		// filter. Panic with what's wrong.
-		panic("score imported, but score.SetPlugin not called")
+	if score == nil { // Then, the user didn't define one.
+		// This is likely caused by use of plugin.Set(p), where 'p' didn't
+		// implement ScorePlugin: return success and score zero.
+		return 0
```
codefromthecrypt, line 64: So, if you use the simple `plugin.Set(plugin)` we have to return something for score. Returning zero here (both status and score). There's no way to avoid this in simple mode because the host will call the function regardless. Only other way is complicated, but exposing a bitflag of functions that could be called.

I think that's too much, so if someone needs to assign score, really they should implement it. If they must not call score, they should not use `plugin.Set(p)` as the docs mention.

scheduler/plugin/plugin.go
```diff
@@ -79,22 +76,6 @@ func NewFromConfig(ctx context.Context, config PluginConfig) (framework.Plugin,
 	}
 }
 
-func readFromURI(ctx context.Context, u string) ([]byte, error) {
-	uri, err := url.ParseRequestURI(u)
```
codefromthecrypt, line 83: the problem was eagerly parsing here (not done in dapr where some logic was copied from). This breaks relative paths like `file://../../foo.wasm`. I noticed this running `make profile`

scheduler/plugin/wasm_test.go
```diff
@@ -0,0 +1,177 @@
...
+			if want, have := tc.expected, bin; !bytes.Equal(want, have) {
+				t.Fatalf("unexpected binary: want %v, have %v", want, have)
+			}
+		})
+	}
+}
+
+func Test_httpGet(t *testing.T) {
+	wasmBinary := wasmMagicNumber
+	wasmBinary = append(wasmBinary, 0x00, 0x00, 0x00, 0x00)
```
codefromthecrypt, line 121: ps I have no idea why we are appending 4 bytes as this was copied from before.. possibly to give some more data to compress?

evacchi: I think it was padded to make it a valid wasmBinary, but this could be misguided. I didn't think of verifying without the padding.

scheduler/plugin/plugin.go
evacchi: ah right. This was not caught by other tests because they were eagerly converting to an absolute `file:///`

codefromthecrypt:
added polish, mostly README

scheduler/plugin/wasm.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// httpGet returns a byte slice of the wasm module found at the given URL, or
+// an error otherwise.
+func httpGet(ctx context.Context, client *http.Client, url string) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Do(req.WithContext(ctx))
```
sanposhiho, line 60: I guess just `req` without `WithContext()` is enough as we're using `NewRequestWithContext` to initialize `req`.

scheduler/plugin/wasm.go
```diff
@@ -0,0 +1,73 @@
...
+	"net/http"
+	"os"
+	"strings"
+)
+
+// getURL parses the URL manually, so that it can resolve relative file paths,
+// such as "file://../../path/to/plugin.wasm"
+func getURL(ctx context.Context, url string) ([]byte, error) {
+	if url == "" {
+		return nil, errors.New("missing url")
```
sanposhiho, line 33: This error is going to appear to users. So, can we make it a bit more fancy like "GuestURL is required in the plugin configuration"?

guest/plugin/plugin.go
```diff
@@ -0,0 +1,44 @@
+package plugin
```
sanposhiho, line 1: license please.

sanposhiho (approved):
Thanks, left small nits.

/approve

scheduler/plugin/wasm.go
codefromthecrypt: yep removed!

scheduler/plugin/wasm.go
codefromthecrypt: I moved the missing one higher up and backfilled a test, so that they are like the others (prefixed `wasm: some error`)

guest/plugin/plugin.go
codefromthecrypt: gotit

codefromthecrypt:
squashed and ready, thanks for the look @sanposhiho!

scheduler/plugin/plugin.go
```diff
@@ -53,9 +51,13 @@ func New(configuration runtime.Object, frameworkHandle framework.Handle) (framew
 // NewFromConfig is like New, except it allows us to explicitly provide the
 // context and configuration of the plugin. This allows flexibility in tests.
 func NewFromConfig(ctx context.Context, config PluginConfig) (framework.Plugin, error) {
-	guestBin, err := readFromURI(ctx, config.GuestURL)
+	url := config.GuestURL
+	if url == "" {
+		return nil, errors.New("wasm: guestURL is required")
```
sanposhiho, line 56: 👍

sanposhiho (approved):
/approve
/lgtm

Thanks!

Issue #72, opened 2023-10-16
Support all extension points


sanposhiho:
- [x] EnqueueExtensions (without hint)
- [ ] EnqueueExtensions (with hint)
- [x] PreFilterExtensions @chansuke 
- [x] PreFilterPlugin
- [x] FilterPlugin
- [x] PreScorePlugin
- [x] ScorePlugin
- [x] ScoreExtensions @Gekko0114 
- [x] PostFilterPlugin @Gekko0114 
- [x] ReservePlugin @utam0k 
- [x] PermitPlugin @kazukousen 
- [x] PreBindPlugin @Gekko0114 
- [x] BindPlugin @Gekko0114 
- [x] PostBindPlugin @Gekko0114 

---

Note that this issue's scope is only the interface support. 
For example, you may think that PostFilterPlugin is useless unless some functions in `framework.Handle` is supported. That might be true, but https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/73 does the support, not this issue.

sanposhiho:
I'll take `ScoreExtensions`.

Gekko0114:
@sanposhiho ,
Can I work on PostFilterPlugin?

sanposhiho:
Yes.

/assign @Gekko0114

Gekko0114:
I'll work on ScoreExtensions as well

sanposhiho:
@Gekko0114 Excellent, assigned it to you.
Sorry for the delay on reviews, I think I'll be able to have time in this weekend.

Gekko0114:
Since implementing extension points is interesting for me, I will work on PreBindPlugin :)
(Please let me know if you have any comments)

Gekko0114:
PreBindPlugin's implementation partially depends on bindPlugin, so I will implment BindPlugin at the same time

Gekko0114:
Thanks for your review!
I will work on PostBindPlugin as well

utam0k:
I'm newer in this field, but I'm interested in contributing to this issue. Can I pick up PreBind?

Gekko0114:
I completed prebind, so remaining item is reserve plugin I guess

utam0k:
@Gekko0114 Wow! You work very hard! Okay, can I pick the Revere plugin? I don't know if this one is suitable for the beginner... 🤔

utam0k:
@sanposhiho Which one is good for the first step?

[...]

PR #75, opened 2023-10-28
support postfilter plugin

Gekko0114:
#### What type of PR is this?
/kind feature
/kind api-change

Support postfilter plugin

Part of #72 

```release-note
Support postfilter plugin
```

```benchstat
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
                                      │  before.txt  │              after.txt               │
                                      │    sec/op    │    sec/op      vs base               │
Example_NodeNumber/Simple/New-4         198.7m ± 79%    222.6m ± 16%        ~ (p=0.310 n=6)
Example_NodeNumber/Simple/Run-4         475.3µ ± 40%    767.4µ ± 41%  +61.46% (p=0.041 n=6)
Example_NodeNumber/Simple_Log/New-4     189.3m ±  0%    209.4m ± 31%  +10.59% (p=0.015 n=6)
Example_NodeNumber/Simple_Log/Run-4     348.7µ ± 31%    411.6µ ± 18%        ~ (p=0.132 n=6)
Example_NodeNumber/Advanced/New-4       217.4m ±  2%    417.0m ± 34%  +91.86% (p=0.002 n=6)
Example_NodeNumber/Advanced/Run-4       79.20µ ±  1%    95.36µ ± 34%  +20.40% (p=0.002 n=6)
Example_NodeNumber/Advanced_Log/New-4   216.7m ±  1%    257.8m ± 39%  +18.94% (p=0.002 n=6)
Example_NodeNumber/Advanced_Log/Run-4   95.40µ ±  2%   131.44µ ± 18%  +37.77% (p=0.002 n=6)
geomean                                 6.213m          8.169m        +31.50%
```


Makefile
```diff
@@ -37,7 +37,7 @@ guest/.tinygo-target.json: scheduler/go.mod
 build-wat: $(wildcard scheduler/test/testdata/*/*.wat)
 	@for f in $^; do \
         wasm=$$(echo $$f | sed -e 's/\.wat/\.wasm/'); \
-		wat2wasm -o $$wasm --debug-names $$f; \
```
Gekko0114, line 40: When I ran original `make build-wat` command, I encountered the error: `The 'path' argument must be of type string. Received undefined.`
Is this command correct?

guest/internal/postfilter/postfilter_exports.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"runtime"
+	"unsafe"
+
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/imports"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/plugin"
+)
+
+// TODO: should postfilter be defined internally?
```
Gekko0114, line 30: I defined postfilter internally same as prefilter. Is it correct?

guest/internal/postfilter/types.go
```diff
@@ -0,0 +1,72 @@
+package postfilter
```
Gekko0114, line 1: This file is copy of prefilter/types.go
I might remove this file later

scheduler/plugin/mask.go
```diff
@@ -49,6 +49,7 @@ func maskInterfaces(plugin *wasmPlugin) (framework.Plugin, error) {
 		iPreBindPlugin |
 		iPostBindPlugin)
 
+	// TODO: ask about PostFilter
```
Gekko0114, line 52: How should I add `iPostFilterPlugin` for this `switch-case` logic?

Makefile
Gekko0114: Seems like CI works fine. So maybe this is because I am using macOS

Gekko0114:
Hi @sanposhiho ,

Though this is a draft, I have some questions.
Could you answer my comments above?
Thanks,

guest/api/types.go
```diff
@@ -97,3 +104,5 @@ type NodeInfo interface {
 
 	Node() proto.Node
 }
+
+type NodeToStatusMap map[string]*Status
```
Gekko0114, line 108: I should define interface here?

scheduler/plugin/mask.go
sanposhiho: No need to add.
When filter is used, postfilter is automatically enabled. I cannot think of any usecase where postfilter needs to be used without filter for now.

https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/2275d37d42e823a564f28bb41f6e9bc3daa3ec88/scheduler/plugin/mask.go#L241-L246

guest/internal/postfilter/postfilter_exports.go
Gekko0114: No. I don't need to define it internally

guest/internal/imports/imports.go
```diff
@@ -29,5 +29,8 @@ func k8sApiNodeName(ptr uint32, limit mem.BufLimit) (len uint32)
 //go:wasmimport k8s.io/api pod
 func k8sApiPod(ptr uint32, limit mem.BufLimit) (len uint32)
 
+//go:wasmimport k8s.io/api nodeToStatusMap
```
sanposhiho, line 32: It should be exported from k8s.io/scheduler since it's a scheduler specific type.

guest/postfilter/postfilter.go
```diff
@@ -0,0 +1,136 @@
...
+	plugin.MustSet(postfilterPlugin)
+}
+
+// prevent unused lint errors (lint is run with normal go).
+var _ func() uint64 = _postfilter
+
+// _postfilter is only exported to the host.
+//
+//export postfilter
+func _postfilter() uint64 { //nolint
```
sanposhiho, line 70: uint32 is enough? See prefilter for example which also returns status.

guest/testdata/cyclestate/main.go
```diff
@@ -109,10 +111,25 @@ func (statePlugin) Filter(state api.CycleState, pod proto.Pod, _ api.NodeInfo) (
...
 }
 
-func (statePlugin) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+func (statePlugin) PostFilter(state api.CycleState, pod proto.Pod, _ api.NodeToStatusMap) (nominatedNodeName string, nominatingMode api.NominatingMode, status *api.Status) {
 	if unsafe.Pointer(pod.Spec()) != unsafe.Pointer(podSpec) {
 		panic("didn't cache pod from filter")
 	}
+	mustNotScoreState(state)
+	if val, ok := state.Read(preFilterStateKey); !ok {
+		panic("didn't propagate state from pre-filter")
```
sanposhiho, line 120: post-filter

guest/testdata/cyclestate/main.go
```diff
@@ -109,10 +111,25 @@ func (statePlugin) Filter(state api.CycleState, pod proto.Pod, _ api.NodeInfo) (
...
+		panic("filter value lost propagating from post-filter")
+	} else {
+		val.(preFilterStateVal)["postfilter"] = struct{}{}
+	}
+	return
+}
+
+func (statePlugin) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	if unsafe.Pointer(pod.Spec()) != unsafe.Pointer(podSpec) {
+		panic("didn't cache pod from post-filter")
```
sanposhiho, line 131: pre-filter

kubernetes/proto/nodetostatus/api.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto2";
```
sanposhiho, line 1: We shouldn't do this. Those protobuf definitions in k8s/proto are all generated ones.

sanposhiho: I don't see any necessity to define this proto in the first place.
NodeToStatusMap contains status, and status contains both reason and status code. But, in a real world, we don't need a reason in PostFilter. (at least, I don't come up with any usecase.)

So, let's simplify here -- use `nodename(string):statuscode(number)` map as nodeToStatusMap in wasm guest side so that we can remove this proto.

sanposhiho (commented):
first round review -- mostly looking good, thanks!

guest/postfilter/postfilter.go
Gekko0114: `(uint64(nominatingMode) << uint64(32)) | uint64(imports.StatusToCode(status))`

Since two values `nominatingMode` and `status` are returned here, I set uint64 as value

guest/internal/imports/imports.go
Gekko0114: Thanks. Fixed

kubernetes/proto/nodetostatus/api.proto
Gekko0114: Thanks. Fixed it. Used json.marshal and json.unmarshal

guest/testdata/cyclestate/main.go
Gekko0114: Added post-filter here

guest/testdata/cyclestate/main.go
Gekko0114: Fixed. BTW prescore is after post-filter. So I thought it should be post-filter. Do I misunderstand something?

Gekko0114:
@sanposhiho ,
Thanks for your comments!
Fixed or replied. PTAL again

guest/testdata/cyclestate/main.go
sanposhiho: After post-filter, no extension point is called.

The scheduling cycle is either
- prefilter → filter --(if no suitable node)--> post-filter → (owari)
- prefilter → filter → prescore → score → ...

guest/testdata/cyclestate/main.go
```diff
@@ -153,5 +170,7 @@ func mustFilterState(state api.CycleState) {
 		panic("didn't propagate state from pre-filter")
 	} else if _, ok = val.(preFilterStateVal)["filter"]; !ok {
 		panic("filter value lost propagating from pre-score")
+	} else if _, ok = val.(preFilterStateVal)["postfilter"]; !ok {
+		panic("postfilter value lost propagating from pre-score")
```
sanposhiho, line 174: I don't think we need this change. As mentioned earlier, no extension point is called after postfilter.

guest/testdata/cyclestate/main.go
```diff
@@ -109,10 +111,25 @@ func (statePlugin) Filter(state api.CycleState, pod proto.Pod, _ api.NodeInfo) (
...
 	if unsafe.Pointer(pod.Spec()) != unsafe.Pointer(podSpec) {
 		panic("didn't cache pod from filter")
 	}
+	mustNotScoreState(state)
+	if val, ok := state.Read(preFilterStateKey); !ok {
+		panic("didn't propagate post-filter state from pre-filter")
+	} else if _, ok = val.(preFilterStateVal)["filter"]; !ok {
+		panic("filter value lost propagating from post-filter")
+	} else {
+		val.(preFilterStateVal)["postfilter"] = struct{}{}
```
sanposhiho, line 124: no need to store something here in preFilterStateVal, no ext point is called post filter.

guest/testdata/cyclestate/main.go
```diff
@@ -109,10 +111,25 @@ func (statePlugin) Filter(state api.CycleState, pod proto.Pod, _ api.NodeInfo) (
...
-func (statePlugin) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+func (statePlugin) PostFilter(state api.CycleState, pod proto.Pod, _ api.NodeToStatusMap) (nominatedNodeName string, nominatingMode api.NominatingMode, status *api.Status) {
 	if unsafe.Pointer(pod.Spec()) != unsafe.Pointer(podSpec) {
 		panic("didn't cache pod from filter")
 	}
+	mustNotScoreState(state)
+	if val, ok := state.Read(preFilterStateKey); !ok {
+		panic("didn't propagate post-filter state from pre-filter")
+	} else if _, ok = val.(preFilterStateVal)["filter"]; !ok {
+		panic("filter value lost propagating from post-filter")
```
sanposhiho, line 122:
```suggestion
		panic("filter value lost propagating from filter")
```

guest/testdata/cyclestate/main.go
```diff
@@ -109,10 +111,25 @@ func (statePlugin) Filter(state api.CycleState, pod proto.Pod, _ api.NodeInfo) (
...
 }
 
-func (statePlugin) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+func (statePlugin) PostFilter(state api.CycleState, pod proto.Pod, _ api.NodeToStatusMap) (nominatedNodeName string, nominatingMode api.NominatingMode, status *api.Status) {
 	if unsafe.Pointer(pod.Spec()) != unsafe.Pointer(podSpec) {
 		panic("didn't cache pod from filter")
 	}
+	mustNotScoreState(state)
+	if val, ok := state.Read(preFilterStateKey); !ok {
+		panic("didn't propagate post-filter state from pre-filter")
```
sanposhiho, line 120:
```suggestion
		panic("didn't propagate state from pre-filter")
```

guest/testdata/cyclestate/main.go
sanposhiho: Also, this cache for pod should always be stored at pre-fiter.
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/6354c6d855a7149042809b6e881bb65d16c31a77/guest/testdata/cyclestate/main.go#L90

guest/api/types.go
```diff
@@ -97,3 +113,8 @@ type NodeInfo interface {
 
 	Node() proto.Node
 }
+
+// PostFilterPlugin use nodeToStatusMap
+type NodeToStatusMap interface {
+	NodeToStatusMap() map[string]int
+}
```
sanposhiho, line 120:
```suggestion
// NodeToStatusMap contains which Node got which status during the scheduling cycle.
type NodeToStatusMap interface {
	// NodeToStatusMap returns an actual map\
	// which is keyed by the node name and valued by the status code.
	NodeToStatusMap() map[string]StatusCode
```

Let's use StatusCode, it tells the meaning of map value implicitly

guest/testdata/cyclestate/main.go
Gekko0114: Fixed it, thanks.

guest/testdata/cyclestate/main.go
Gekko0114: > Also, this cache for pod should always be stored at pre-fiter.

Fixed it, but if I misunderstood, please let me know.

Gekko0114:
Thanks for your review, @sanposhiho !
Fixed it, PTAL again.

sanposhiho (approved):
Perfection, it's a huge step, thanks!

/lgtm
/approve

guest/api/types.go
```diff
@@ -97,3 +113,8 @@ type NodeInfo interface {
 
 	Node() proto.Node
 }
+
+// PostFilterPlugin use nodeToStatusMap
+type NodeToStatusMap interface {
+	Map() map[string]StatusCode
```
sanposhiho, line 119: It's better to be:

```suggestion
// PostFilterPlugin use nodeToStatusMap
type NodeToStatus interface {
	Map() map[string]StatusCode
```

---

Also, you missed the comment change suggested on https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/75#discussion_r1403489824

sanposhiho (commented):
one nit

guest/api/types.go
Gekko0114: Thanks, fixed this comment and your previous comment.

Gekko0114:
@sanposhiho ,
Thanks, fixed it. PTAL

sanposhiho (approved):
/lgtm
/approve
/unhold

PR #76, opened 2023-11-23
support scoreExtensions

Gekko0114:
#### What type of PR is this?

/kind feature
/kind api-change

Support score Extensions

Part of #72 

```release-note
support score Extensions
```

```benchstat

```


Gekko0114:
Hi @sanposhiho,

I've created a PR for scoreExtensions. PTAL, though I haven't implemented cache of each node's score.

guest/api/types.go
```diff
@@ -91,9 +91,20 @@ type ScorePlugin interface {
...
+
 type NodeInfo interface {
 	// Metadata is a convenience that triggers Get.
 	proto.Metadata
 
 	Node() proto.Node
 }
+
+type NodeScoreList interface {
+	NodeScoreList() map[string]int
```
sanposhiho, line 109: It's also a similar one, let's change like:

```suggestion
type NodeScore interface {
	// Map returns a map
	// which is keyed by the node name and valued by the score.
	Map() map[string]int
```

sanposhiho: It's a general suggestion, when you use map in interface, it's always better to say explicitly what is key and what is value.

guest/scoreextensions/scoreextensions.go
```diff
@@ -0,0 +1,102 @@
...
+//	}
+//
+//	type scoreExtensionsPlugin struct{}
+//
+//	func (scoreExtensionsPlugin) NormalizeScore(state api.CycleState, pod api.Pod, nodeScoreList map[string]int) (map[string]int, *api.Status) {
+//		panic("implement me")
+//	}
+func SetPlugin(scoreExtensions api.ScoreExtensions) {
+	if scoreExtensions == nil {
+		panic("nil scorePlugin")
```
sanposhiho, line 51:
```suggestion
		panic("nil scoreExtensions")
```

sanposhiho (commented):
/lgtm

One nit picked, lgtm otherwise

guest/api/types.go
Gekko0114: Thanks, fixed it

Gekko0114:
Hi @sanposhiho ,
Thanks for your review. Fixed it, PTAL

Issue #73, opened 2023-10-16
Support `framework.Handle`


sanposhiho:
The scheduler exports `framework.Handle` to the plugins so that they can operate something on the framework side.

https://github.com/kubernetes/kubernetes/blob/add482a2d31db938f3dca87920313db9d1780cf5/pkg/scheduler/framework/interface.go#L601

Some functions don't need to be implemented though, we should support the common ones at least:
- [ ] SnapshotSharedLister
- [x] EventRecorder @Gekko0114 https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/83
- [ ] RunFilterPluginsWithNominatedPods @chansuke 
- [ ] GetWaitingPod @chansuke 
- [x] RejectWaitingPod @Gekko0114 https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/98

sanposhiho:
SnapshotSharedLister might not be needed if https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/71 is complited

Gekko0114:
@sanposhiho 
I would like to work on another one, do you have any recommendation? (which item has higher priority?)

sanposhiho:
I would say `RunFilterPluginsWithNominatedPods` is the one.

Gekko0114:
Thanks, then I will work on RunFilterPluginsWithNominatedPods next

Gekko0114:
Since RunFilterPlugins is complicated, I will work on GetWaitingPod instead

chansuke:
@sanposhiho @Gekko0114
I would like to contribute to this issue.
I'm interested in GetWaitingPod, but could you please share the priority and current status? :pray:
Or any recommendations?

chansuke:
@Gekko0114 
Thank you so much for your comment.
I'll take a look.

sanposhiho:
/assign @chansuke 

Thanks!

sanposhiho:
/assign

Will pick up `SnapshotSharedLister`.

sanposhiho:
/priority next-release
Not all, but at least, `SnapshotSharedLister` is.

chansuke:
@sanposhiho @Gekko0114
I'd like to work on `RunFilterPluginsWithNominatedPods` if no one is working.

sanposhiho:
@chansuke 👍

PR #83, opened 2023-12-06
Support EventRecorder

Gekko0114:
#### What type of PR is this?
/kind feature

Optionally add one or more of the following kinds if applicable:
/kind api-change

Support EventRecorder

Part of #73 

```release-note
Support EventRecorder
```

```benchstat
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
                                      │  before.txt  │              after.txt               │
                                      │    sec/op    │    sec/op      vs base               │
Example_NodeNumber/Simple/New-4         188.5m ± 10%    200.2m ±  8%        ~ (p=0.485 n=6)
Example_NodeNumber/Simple/Run-4         616.8µ ± 71%    322.9µ ± 24%  -47.64% (p=0.026 n=6)
Example_NodeNumber/Simple_Log/New-4     299.5m ± 31%    210.5m ± 29%        ~ (p=0.093 n=6)
Example_NodeNumber/Simple_Log/Run-4     393.9µ ± 28%    305.7µ ± 18%  -22.38% (p=0.015 n=6)
Example_NodeNumber/Advanced/New-4       258.5m ± 45%    248.6m ± 65%        ~ (p=1.000 n=6)
Example_NodeNumber/Advanced/Run-4       94.82µ ± 33%   129.47µ ± 28%  +36.54% (p=0.015 n=6)
Example_NodeNumber/Advanced_Log/New-4   258.6m ± 13%    299.3m ± 21%        ~ (p=0.132 n=6)
Example_NodeNumber/Advanced_Log/Run-4   99.73µ ± 21%   126.66µ ± 10%  +27.01% (p=0.002 n=6)
geomean                                 7.367m          6.890m         -6.48%
```


guest/handle/api/handle.go
```diff
@@ -0,0 +1,29 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package api
+
+import (
+	internalproto "sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/proto"
+)
+
```
Gekko0114, line 22: add comment

Gekko0114:
add test case calling eventRecorder in wasm

examples/advanced/plugin/plugin_test.go
```diff
@@ -144,6 +144,18 @@ func (t testPod) GetNamespace() string {
 	return ""
 }
 
+func (t testPod) GetApiVersion() string {
+	return ""
+}
+
+func (t testPod) GetKind() string {
```
Gekko0114, line 151: ABI doesn't include kind and resouce version. So we return static values here

Gekko0114:
/assign @sanposhiho ,

Created a PR. PTAL. Thanks,

sanposhiho:
Sorry for the delay, will review it in this weekend **hopefully**.

Gekko0114:
If I need to rebase before reviewing, please let me know

examples/advanced/main.go
```diff
@@ -36,19 +38,44 @@ import (
...
+	// Instead of using `plugin.Set`, this configures only the interfaces
+	// implemented by the plugin. The Wasm host only calls functions imported,
+	// so this prevents additional overhead.
+	enqueue.SetPlugin(func(h handleapi.Handle) api.EnqueueExtensions {
+		p := pluginInitializer(h)
+		plugin, ok := p.(api.EnqueueExtensions)
+		if ok {
+			return plugin
+		}
+		panic("pluginInitializer did not return an api.EnqueueExtensions")
```
sanposhiho, line 50:
```suggestion
		return p.(api.EnqueueExtensions)
```

It panics anyway if p doesn't implement EnqueueExtensions.

examples/advanced/plugin/plugin.go
```diff
@@ -73,10 +75,18 @@ func (pl *NodeNumber) EventsToRegister() []api.ClusterEvent {
 
 // PreScore implements api.PreScorePlugin
 func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	var recorder handleapi.EventRecorder
+	h := pl.handle
+	if h != nil {
+		recorder = h.EventRecorder()
+	}
```
sanposhiho, line 82: `h` shouldn't be nil, right?

examples/nodenumber/main.go
```diff
@@ -82,10 +84,19 @@ func (pl *NodeNumber) EventsToRegister() []api.ClusterEvent {
 
 // PreScore implements api.PreScorePlugin
 func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	var recorder handleapi.EventRecorder
+	h := pl.handle
+	if h != nil {
```
sanposhiho, line 89: ditto

examples/advanced/main.go
```diff
@@ -36,19 +38,44 @@ import (
 // main is compiled to an exported Wasm function named "_start", called by the
 // Wasm scheduler plugin during initialization.
 func main() {
+	// Instead of using `plugin.Set`, this configures only the interfaces
+	// implemented by the plugin. The Wasm host only calls functions imported,
+	// so this prevents additional overhead.
```
sanposhiho, line 43: Wouldn't it be the same if we use `plugin.Set`? since `plugin.Set` does the interface checks.

examples/advanced/main.go
```diff
@@ -36,19 +38,44 @@ import (
...
 		panic(err)
 	}
-	// Instead of using `plugin.Set`, this configures only the interfaces
-	// implemented by the plugin. The Wasm host only calls functions imported,
-	// so this prevents additional overhead.
-	enqueue.SetPlugin(plugin)
-	prescore.SetPlugin(plugin)
-	score.SetPlugin(plugin)
+	return plugin
 }
```
sanposhiho, line 81: It doesn't look great. 

We have two alternative choices:
- change `Set` functions to accept `func (klog klog.Klog, jsonConfig []byte, h handleapi.Handle)`
- change `New` function of this plugin to only have `h handleapi.Handle` parameter, and change it to do klog.Get() and config.Get() in it.

I prefer the first option. Actually, the scheduling framework's initializer type contains configuration in the second parameters, so it's not weird to add it to our plugin initializer type as well. 
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/runtime/registry.go#L31
Also, for logger, we may want to add it to handleapi.Handle.

guest/handle/eventrecorder.go
```diff
@@ -0,0 +1,94 @@
...
+		RelatedReference:   rel,
+		Eventtype:          eventtype,
+		Reason:             reason,
+		Action:             action,
+		Note:               note,
+	}
+	e.EventfFn(msg)
+}
+
+func ConvertToObjRef(obj internalproto.KObject) ObjectReference {
```
sanposhiho, line 62: no need to make it public function?

sanposhiho (commented):
I don't have enough time going thru all, but left some comments.

examples/advanced/main.go
Gekko0114: I see, you're right. When using `plugin.Set`, it also performs an interface check before calling each plugin. Therefore, it seems there is no difference between using `SetPlugin` and `Set`.
I will remove comments here.

examples/advanced/main.go
Gekko0114: Sure, therefore I will change pluginInitializer to  be `func (klog klog.Klog, jsonConfig []byte, h handleapi.Handle)`.

> Also, for logger, we may want to add it to handleapi.Handle.

Then in the near future, we might remove `klog` and integrate it into `handle`.

examples/advanced/main.go
```diff
@@ -26,29 +26,40 @@ import (
...
```
sanposhiho, line 65: Ah, so we still need `pluginInitializer` because `plugin.New` returns error. I guess we can change `SetPlugin` to accept the function `func (klog klog.Klog, jsonConfig []byte, h handleapi.Handle) (api.Plugin, error)`?

examples/advanced/plugin/plugin.go
```diff
@@ -73,10 +75,16 @@ func (pl *NodeNumber) EventsToRegister() []api.ClusterEvent {
...
 // PreScore implements api.PreScorePlugin
 func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	var recorder handleapi.EventRecorder
+	h := pl.handle
+	recorder = h.EventRecorder()
 	pl.klog.InfoS("execute PreScore on NodeNumber plugin", "pod", klog.KObj(pod))
 
 	podnum, ok := lastNumber(pod.Spec().GetNodeName())
 	if !ok {
+		if recorder != nil {
```
sanposhiho, line 85: Could `recorder` be nil?

guest/filter/filter.go
```diff
@@ -46,11 +51,12 @@ var filter api.FilterPlugin
 //	}
 //
 // Note: If you need state, you can assign it with prefilter.SetPlugin.
-func SetPlugin(filterPlugin api.FilterPlugin) {
-	if filterPlugin == nil {
+func SetPlugin(pluginInitializer func(klog klogapi.Klog, jsonConfig []byte, h handleapi.Handle) api.FilterPlugin) {
```
sanposhiho, line 54: I think we should define the type of initializer function somewhere like the scheduling framework does.
And, let's name the type PluginFactory, following the framework.
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/runtime/registry.go#L31

examples/advanced/main.go
sanposhiho: And then, do we still need to use `SetPlugin` here?

examples/advanced/main.go
Gekko0114: > And then we can just pass plugin.New to SetPlugin.

I didn't notice this point.. Thanks, fixed it

guest/filter/filter.go
Gekko0114: Thanks, fixed it

Gekko0114:
@sanposhiho ,

Thanks for your review. PTAL again

examples/advanced/plugin/plugin.go
```diff
@@ -73,10 +75,14 @@ func (pl *NodeNumber) EventsToRegister() []api.ClusterEvent {
 
 // PreScore implements api.PreScorePlugin
 func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	var recorder handleapi.EventRecorder
+	h := pl.handle
+	recorder = h.EventRecorder()
```
sanposhiho, line 80:
```suggestion
	h := pl.handle
	recorder := h.EventRecorder()
```

examples/nodenumber/main.go
```diff
@@ -82,10 +87,15 @@ func (pl *NodeNumber) EventsToRegister() []api.ClusterEvent {
 
 // PreScore implements api.PreScorePlugin
 func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ proto.NodeList) *api.Status {
+	var recorder handleapi.EventRecorder
+	h := pl.handle
+	recorder = h.EventRecorder()
```
sanposhiho, line 92: ditto

examples/nodenumber/main.go
```diff
@@ -39,7 +41,9 @@ func main() {
 		}
 		klog.Info("NodeNumberArgs is successfully applied")
 	}
-	plugin.Set(&NodeNumber{reverse: args.Reverse})
+	plugin.Set(func(klog klogapi.Klog, jsonConfig []byte, h handleapi.Handle) (api.Plugin, error) {
+		return &NodeNumber{reverse: args.Reverse, handle: h}, nil
+	})
```
sanposhiho, line 46: nit: Can you create `New` function like `/examples/advanced/` does?

scheduler/plugin/host.go
```diff
@@ -48,6 +49,8 @@ const (
 	k8sSchedulerResultNominatedNodeName   = "result.nominated_node_name"
 	k8sSchedulerResultStatusReason        = "result.status_reason"
 	k8sSchedulerResultNormalizedScoreList = "result.normalized_score_list"
+	k8sHandle                             = "k8s.io/handle"
+	k8sHandleEventRecorderEventf          = "eventrecorder.eventf"
```
sanposhiho, line 53: This should be in `k8sScheduler` module since `framework.Handle` is scheduler specific stuff rather than something in k8s in general.

scheduler/plugin/plugin_test.go
```diff
@@ -1350,6 +1352,52 @@ func TestPostBind(t *testing.T) {
...
+			guestURL:    test.URLExampleNodeNumber,
+			pod:         test.PodSmall,
+			expectedMsg: "PreScore not match lastNumber Skip ",
+		},
+		{
+			name:        "call each plugin's Set",
+			guestURL:    test.URLExampleAdvanced,
+			pod:         test.PodSmall,
+			expectedMsg: "PreScore not match lastNumber Skip ",
+		},
```
sanposhiho, line 1374: What're those test case titles meaning? (what's plugin.Set or plugin's Set they're referring to?)

Gekko0114: Sorry, these are not clear. Changed titles.
Does it make sense?

Gekko0114:
@sanposhiho ,
Thanks for your review. Fixed it.
PTAL when you have time (it's not urgent)

examples/advanced/main.go
```diff
@@ -26,29 +26,15 @@ import (
...
 
 // main is compiled to an exported Wasm function named "_start", called by the
 // Wasm scheduler plugin during initialization.
 func main() {
-	// The plugin package uses only normal Go code, which allows it to be
-	// unit testable via `tinygo test -target=wasi` as well normal `go test`.
-	//
-	// The real implementations, such as `config.Get()` use Wasm host functions
-	// (go:wasmimport), which cannot be tested with `tinygo test -target=wasi`.
-	plugin, err := plugin.New(klog.Get(), config.Get())
```
codefromthecrypt, line 44: this line prevented the direct plumbing of things that would interfere with tests, specifically binding to `klog.Get(), config.Get()`. If something is broke, look here. Advanced was how to show how to unit test inside tinygo, and likely this is still needed.

examples/nodenumber/main.go
```diff
@@ -25,21 +25,28 @@ import (
...
 // main is compiled to a WebAssembly function named "_start", called by the
 // wasm scheduler plugin during initialization.
 func main() {
-	var args nodeNumberArgs
 	if jsonConfig := config.Get(); jsonConfig != nil {
 		if err := json.Unmarshal(jsonConfig, &args); err != nil {
 			panic(fmt.Errorf("decode arg into NodeNumberArgs: %w", err))
 		}
 		klog.Info("NodeNumberArgs is successfully applied")
 	}
```
sanposhiho, line 44: This should be moved into `New`.

sanposhiho: Because jsonConfig is passed as a parameter of `New`

scheduler/plugin/plugin_test.go
sanposhiho: yep

guest/handle/api/handle.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	internalproto "sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/proto"
+	klogapi "sigs.k8s.io/kube-scheduler-wasm-extension/guest/klog/api"
+)
+
+// Handle provides data and some tools that plugins can use.
+//
+// This contains functions like framework.Handle.
+type Handle interface {
```
codefromthecrypt, line 28: In wasm I don't think we need to introduce the handle type, which is similar to a New function. It may be better to not introduce that vocab and instead only define EventRecorder similar to KLog here. In other words I think probably delete handle to flatten the types.

Also, you want to be sure that the api can be useable without depending on implementation,and tests here and in the advanced example will prove that. See. https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/guest/klog/api/klog.go https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/guest/klog/api/klog_test.go and try to follow the style as it may prevent some pitfalls.

guest/handle/api/handle.go
```diff
@@ -0,0 +1,52 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package api
+
+import (
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	internalproto "sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/proto"
```
codefromthecrypt, line 21: If this is really api, you probalby wouldn't be depending on internal/proto. I think this will cause some package dependency problems. I would expect sigs.k8s.io/kube-scheduler-wasm-extension/guest/api/proto" and only the implementation to use the proto types. You can look at the previous functions in various places for examples of this.

guest/handle/api/handle.go
Gekko0114: > In wasm I don't think we need to introduce the handle type, which is similar to a New function. It may be better to not introduce that vocab and instead only define EventRecorder similar to KLog here. In other words I think probably delete handle to flatten the types.

@sanposhiho,
What do you think about this suggestion?
It seems reasonable as the handle on the guest side is more of a formality, so it's possible to call EventRecorder directly without the handle.
However, if we make this change, the pluginFactory will become unnecessary, which means significant changes to the current implementation of the PR.
I prefer current implementation because it's similar to current kubernetes implementation itself.

examples/advanced/main.go
Gekko0114: @codefromthecrypt 
Thanks for your comments!
Just clarify, shouldn't I remove this line?
In my implementation, klog.Get() and config.Get() are called within plugin.go, is it not good implementaiton?

guest/handle/api/handle.go
codefromthecrypt: I know you didn't ask me, but I think there is also a RATIONALE.md here which mentions for simplicitly we don't follow exactly upstream SDK, it also improves the performance of the plugin to not have layers. Following the prior patterns of simple binding of types needed defined here, is a good way to be consistent forward. If you want to go back and make this for the sake of it same as the main SDK, maybe do it everywhere so that  it isn't inconsistent. However, the problems solved will be re-introduced, so update the RATIONALE to say why you chose to revert.

codefromthecrypt: TL'DR' is study existing practice and the RATIONALE files which took a long time to work out. If you disagree, then you can change, just make sure you make the whole thing consistent with that change. The RATIONALE was made so that I could step away and lessons remain.

sanposhiho: > simplicitly we don't follow exactly upstream SDK, it also improves the performance and testability of the plugin to not have layers. 

Based on this merit, I thumb up for the suggestion from @codefromthecrypt, for now.
@Gekko0114 's point does also make sense though, we can consider taking such steps that make wasm SDK similar to the original scheduler SDK **later**, along with all the other interfaces + combatting problems introduced by them like @codefromthecrypt suggested, because it takes additional effort for us while being less important.

examples/advanced/main.go
Gekko0114: Thanks for your feedback! Reverted this line.

guest/handle/api/handle.go
Gekko0114: Sure, thanks @sanposhiho  and @codefromthecrypt for your suggestions.
Removed handle and added tests for eventrecorder.

Gekko0114:
Hi @sanposhiho @codefromthecrypt 
Thanks for your suggestions. Based on the discussion, I removed handle. PTAL when you have time.

sanposhiho:
I'm curious to see the benchmark result, can you update your PR description, following the instruction:
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/73a62c8e1abe3e0efe6c523610b3cc270cdcd1f0/.github/PULL_REQUEST_TEMPLATE.md#L53-L65

examples/nodenumber/main.go
```diff
@@ -25,21 +25,32 @@ import (
 	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
 	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api/proto"
 	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/config"
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/eventrecorder"
 	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/klog"
+	klogapi "sigs.k8s.io/kube-scheduler-wasm-extension/guest/klog/api"
 	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/plugin"
 )
 
+var args nodeNumberArgs
```
sanposhiho, line 34: nit: Does it need to be defined here?

internal/e2e/scheduler/scheduler_test.go
sanposhiho, line 1: Can we improve `testExample_NodeNumber` to test event recorders? There's no e2e test for this enhancement for now.

Gekko0114:
@sanposhiho,
Thanks for your review. PTAL again when you have time.

sanposhiho (approved):
/lgtm
/approve

PR #86, opened 2023-12-17
fix: use correct guest in binding cycles

sanposhiho:
#### What type of PR is this?

/kind bug

We have to use guests in guestPool.binding during binding cycles.

Fixes #

```release-note

```

```benchstat

```


Gekko0114:
/lgtm

Thanks!

PR #93, opened 2024-01-08
Support Permit Plugin

kazukousen:
#### What type of PR is this?

/kind feature
/kind api-change

Support Permit Plugin

Part of #72 

`permit` must return `timeout` in addition to the `statusCode`. The Wasm (specification 1) allows only one return value to be defined, and the maximum size is limited to i64. But the statusCode has already consumed i32. Packing the remaining i32, as in the implementation of score, is difficult because timeout cannot be reduced to a size smaller than i64.

Therefore, the one of remaining solutions is to add a host callback for the timeout, Guest passes a timeout to an imported Host function. This will only add hostcall overhead.

So the suggestion would add `set_timeout` to the ABI.

```release-note
Support permit plugin
```

```benchstat

```


Gekko0114:
Overall it looks good! Thank you.
I will add approve label after passing CI checks.

@sanposhiho,
Seems that without your approval, this PR can't run CI checks. Could you take a look?

Gekko0114 (commented):
nit:
could you update these two main.wasm files and run `make test` again? Maybe you need to update some of test cases.

https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/tree/main/examples/advanced
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/tree/main/examples/nodenumber

kazukousen:
Thank you for your kind review. Fixed it 1aa0020. PTAL when you have time.

Gekko0114:
Please fix ci failures.

kazukousen:
@Gekko0114 Sorry, Forced push and then fixed examples/advanced/main.wasm only at 976939b .
The change should not affect examples/nodenumber/main.wasm .

I guess that main.wasm is not synchronized with main.go because [the test](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/73a62c8e1abe3e0efe6c523610b3cc270cdcd1f0/scheduler/plugin/plugin_test.go#L215-L218) expects `FilterPlugin` and `ScorePlugin` interfaces, but the main.go implementation lacks the `FilterPlugin` interface.
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/actions/runs/7450036829/job/20343211771

kazukousen:
Through conversations with @Gekko0114 in Slack, I've got a handle on what's going on.
It seems we should have updated our wasm build and the mask test when ReservePlugin and BindPlugin were added.
be83292

I suppose the permanent solution is to ensure that [the CI](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/73a62c8e1abe3e0efe6c523610b3cc270cdcd1f0/.github/workflows/testdata.yaml#L54) fails when there's a discrepancy.

kazukousen:
@Gekko0114 thanks, Fixed it c0fefd0. PTAL

guest/api/types.go
```diff
@@ -122,6 +126,13 @@ type ReservePlugin interface {
 	Unreserve(state CycleState, p proto.Pod, nodeName string)
 }
 
+// PermitPlugin is a WebAssembly implementation of framework.PermitPlugin.
+type PermitPlugin interface {
+	Plugin
+
+	Permit(state CycleState, p proto.Pod, nodeName string) (*Status, time.Duration)
```
sanposhiho, line 133: Document `time.Duration` should be in the range of uint32 and otherwise the guest panic (?).

guest/testdata/cyclestate/main.go
```diff
@@ -186,6 +189,10 @@ func (statePlugin) Reserve(state api.CycleState, pod proto.Pod, nodeName string)
 func (statePlugin) Unreserve(state api.CycleState, pod proto.Pod, nodeName string) {
 	mustFilterState(state)
 }
+func (statePlugin) Permit(state api.CycleState, pod proto.Pod, nodeName string) (status *api.Status, timeout time.Duration) {
+	// Actually, it is not called from the cycle test, but it needs for Permit plugin.
```
sanposhiho, line 193: Why is it not called? In other words, can we change the test and this `Permit` to make sure the cycle state is propagated to Permit correctly?

guest/api/types.go
kazukousen: We need to constrain the range of `time.Duration` as you mentioned.
So the host call `set_timeout` that I first proposed in this PR is no longer needed, right?

guest/testdata/cyclestate/main.go
kazukousen: Thanks,
I guess it makes sense to verify that the `preFilterStateKey` should be propagated on Permit since the existing _cycle state_ test implementation always calls prefilter, right?

guest/api/types.go
Gekko0114: I thought current implementation is fine because timeout is being passed via linear memory instead of stack. Only status is being passed via stack. So we don't need to restrict the value of timeout in uint32 range. Do I misunderstand?

guest/testdata/cyclestate/main.go
sanposhiho: Yup, that'd be nice.

guest/api/types.go
sanposhiho: > So we don't need to restrict the value of timeout in uint32 range. Do I misunderstand?

We do. `Int64ToPtr` wouldn't behave as user's intention when timeout value is too big. (see: https://go.dev/play/p/tDCFt2Yu0Td)

> So the host call set_timeout that I first proposed in this PR is no longer needed, right?
I'd like to pass the statusCode and time.Duration to host by packing them into a single i64 return value.

LGTM.

kazukousen:
Sorry for the delay. I will work on it this weekend.

kazukousen:
Sorry for the delay.
Unfortunately, I'm currently unable to commit the necessary time to complete this task.
With respect to the community and to prevent any further delays on the project, I think it's best to withdraw the PR and unassign myself from the task.
I sincerely apologize for any inconvenience this may have caused and appreciate your understanding.
Thanks,

sanposhiho:
No worries, appreciate for your work so far.
I'll take over the remaining part 👍

PR #98, opened 2024-01-22
Support RejectWaitingPod

Gekko0114:
#### What type of PR is this?
/kind feature
/kind api-change

Support RejectWaitingPod

Part of #73 

```release-note
Support RejectWaitingPod
```

```benchstat
```


scheduler/plugin/mem.go
```diff
@@ -73,3 +73,9 @@ func writeStringIfUnderLimit(mem wazeroapi.Memory, v string, buf uint32, bufLimi
 	mem.WriteString(buf, v)
 	return vLen
 }
+
+func writeUint64(mem wazeroapi.Memory, v uint64, buf uint32, bufLimit bufLimit) int {
+	size := 8
```
Gekko0114, line 78: Size can be removed

Gekko0114:
/assign @sanposhiho 
Could you review it when you have time? It is not urgent.

sanposhiho:
@Gekko0114 Can we have a test in /internal/e2e as well

examples/advanced/plugin/plugin.go
```diff
@@ -109,6 +110,10 @@ func (pl *NodeNumber) Score(state api.CycleState, pod proto.Pod, nodeName string
 	if match {
 		return 10, nil
 	}
+	// Call RejectWaitingPod for e2e test.
+	if pod != nil {
+		handle.RejectWaitingPod(pod.GetUid())
+	}
```
sanposhiho, line 116: Well, it's not a practical example of using `RejectWaitingPod`.
`RejectWaitingPod` is often used with `Permit` extension point, which any of our examples haven't implemented yet.
You can check https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/coscheduling/coscheduling.go for better example how people use it with Permit generally.

So, it's a general problem that we only have basic plugins, and it's better not to put various usecases into those two basic example plugins.
I'd rather enrich our examples more for many aspects (https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/101).

@Gekko0114 Can you once remove https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/pull/98/commits/713c9b3d1cc1ee90949975112b45f182edec7e77 from this PR and work on having a new example for a new e2e in another PR?

sanposhiho (commented):
/lgtm
/approve

PR #99, opened 2024-01-24
Fix minor bug about %v

Gekko0114:
#### What type of PR is this?
/kind bug

We can't pass specific integer to %v. We get the error below.
!v(PANIC=String method: runtime error: index out of range [2147483647] with length 6)

```release-note
NONE
```

```benchstat

```


sanposhiho (commented):
/lgtm
/approve

PR #116, opened 2024-08-23
fix: repear scheduler_perf to run correctly

sanposhiho:
#### What type of PR is this?

/kind bug

**1. Fix a bug with NormalizedScore**

Currently, when the guest doesn't implement NormalizedScore, the wasm plugin (scheduler side) errors out, which makes scheduler_perf fail right now.
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/scheduler/plugin/plugin.go#L341-L343

We should restore the node list when the guest returns nil.

**2. Create a wasm for scheduler_perf**

We're using examples/nodenumber in scheduler_perf, but examples/nodenumber uses many additional features (klog and event recorder) which creates additional overhead.
With scheduler_perf, we just want to compare the performance difference between (default scheduler,) wasm plugin and extender under the same condition (running Score() or `/priorities` which only has almost the same logic inside).
This PR creates the simplest nodenumber wasm plugin to be used in scheduler_perf for this purpose. 

We can add more wasm plugins per test case later. (e.g., the one that uses logging a lot, etc)

Fixes #

```release-note

```

```benchstat

```


scheduler/plugin/guest.go
```diff
@@ -220,6 +220,11 @@ func (g *guest) normalizeScore(ctx context.Context) (framework.NodeScoreList, *f
 	statusCode := int32(callStack[0])
 	statusReason := paramsFromContext(ctx).resultStatusReason
 	normalizedScoreList := paramsFromContext(ctx).resultNormalizedScoreList
+	if len(normalizedScoreList) == 0 {
```
Gekko0114, line 223: nit: What if the length of `normalizedScoreList` is not equal to the length of `nodeScoreList` ?

Gekko0114 (approved):
Overall looks good, thanks so much for fixing this (I remembered that I had implemented scoreExtensions for wasm plugin)
Left one comment.

scheduler/plugin/guest.go
sanposhiho: We return the error:
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/scheduler/plugin/plugin.go#L341-L343

We shouldn't restore the score list in that case because if the length is unmatched, that's the bug in the guest. (The guest filled in an invalid node score list) 
We should error out, rather than hiding the bug.

And, what we're trying to do here is not hide the bug. When the guest doesn't implement the normalizescore, we just return without doing anything, which results in an empty `resultNormalizedScoreList`, and hence fail.

https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/main/guest/scoreextensions/scoreextensions.go#L64-L68

Gekko0114: > what we're trying to do here is not hide the bug. When the guest doesn't implement the normalizescore, we just return without doing anything, which results in an empty resultNormalizedScoreList, and hence fail.

Make sense. Thank you so much for your explanation!

PR #121, opened 2024-09-14
Support `GetWaitingPod`

chansuke:
#### What type of PR is this?

Adding `GetWaitingPod` of `framework.Handle` 

Partof #73

```release-note
Support GetWaitingPod
```

```benchstat

```


chansuke:
searching [Test_guestPool_bindingCycles](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/actions/runs/11417090923/job/31769022792?pr=121) error

guest/handle/imports.go
```diff
@@ -20,5 +20,11 @@ package handle
 
 import "sigs.k8s.io/kube-scheduler-wasm-extension/guest/internal/mem"
 
+//go:wasmimport k8s.io/scheduler handle.get_waiting_pod
```
sanposhiho, line 23:
```suggestion
//go:wasmimport k8s.io/scheduler handle.allow_waiting_pod
```

sanposhiho: Actually, I don't see the implementation of `allow_waiting_pod` in the host side. 
Do you want to keep the scope of this PR as only getWaitingPod? Or do you want to include that in the PR too? Either is fine to me.

scheduler/plugin/host_test.go
```diff
@@ -96,39 +95,31 @@ func initKlog(t *testing.T, buf *bytes.Buffer) {
 	klog.SetOutput(buf)
 }
 
-func Test_k8sHandleEventRecorderEventFn(t *testing.T) {
```
sanposhiho, line 99: why is this test removed?

chansuke: Thank you for your comment.
I was checking the behavior of test code in `scheduler/plugin/host_test.go` and forgot to restore the `Test_k8sHandleEventRecorderEventFn`

guest/handle/imports.go
chansuke: I'd like to keep the focus of this PR limited to `getWaitingPod`

sanposhiho: If so, please remove all the changes for `allow_waiting_pod`.

guest/api/types.go
```diff
@@ -205,3 +205,6 @@ type NodeScore interface {
 	// which is keyed by the node name and valued by the score.
 	Map() map[string]int
 }
+
+// WaitingPod represents a pod currently waiting in the permit phase.
+type WaitingPod interface{}
```
sanposhiho, line 210: You can keep previous `GetPod` and `Reject` because they're implemented.

guest/handle/handle.go
```diff
@@ -34,3 +42,26 @@ func RejectWaitingPod(uid string) bool {
 	runtime.KeepAlive(uid)
 	return wasmBool == 1
 }
+
+func GetWaitingPod(uid string) guestapi.WaitingPod {
+	ptr, size := mem.StringToPtr(uid)
+
+	// Wrap to avoid TinyGo 0.28: cannot use an exported function as value
+	podBytes := mem.SendAndGetPodBytes(ptr, size, func(input_ptr, input_size, ptr uint32, limit mem.BufLimit) {
```
sanposhiho, line 50: any reason why we have to fetch pod in a different way from others, with a new function `SendAndGetPodBytes`?

e.g., when we fetch node, we do like this:
https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/b7b0d94b38c573aa707b11970d5137ba51f5f431/guest/internal/imports/host.go#L73-L77

chansuke: Thank you for your information. There's no reason to create a new function. I will update this.

guest/internal/mem/mem.go
```diff
@@ -103,3 +103,24 @@ func SendAndGetUint64(input_ptr uint32, input_size uint32, fn func(input_ptr, in
 	fn(input_ptr, input_size, uint32(readBufPtr), readBufLimit)
 	return binary.LittleEndian.Uint64(readBuf)
 }
+
+func SendAndGetPodBytes(input_ptr uint32, input_size uint32, fn func(input_ptr, input_size, ptr uint32, limit BufLimit)) []byte {
```
sanposhiho, line 107: Is it still used?

scheduler/plugin/plugin_test.go
```diff
@@ -1765,6 +1766,87 @@ func TestRejectWaitingPod(t *testing.T) {
 	}
 }
 
+func TestGetWaitingPod(t *testing.T) {
+	tests := []struct {
+		name               string
+		guestURL           string
+		pod                *v1.Pod
+		args               []string
+		expectedUID        types.UID
```
sanposhiho, line 1775: where do we use it?

scheduler/plugin/plugin_test.go
```diff
@@ -1807,3 +1889,56 @@ func requireError(t *testing.T, err error, expectedError string) {
...
+
+func makeTestWaitingPod(pod *v1.Pod, plugins map[string]*time.Timer) framework.WaitingPod {
+	return &waitingPod{
+		pod:            pod,
+		pendingPlugins: plugins,
+	}
+}
+
+// comparePods compares the UIDs of two v1.Pod objects
+func comparePods(pod1, pod2 *v1.Pod) bool {
```
sanposhiho, line 1939: why do we need to create this function? can't we somehow just use the common comparing functions such as reflect.DeepEqual etc?

scheduler/test/testdata.go
```diff
@@ -143,6 +146,54 @@ var PodForHandleTest = &v1.Pod{
...
+	wp.mu.Lock()
+	defer wp.mu.Unlock()
+}
+
+// Create the test WaitingPod
+var WaitingPod = &waitingPod{
+	pod:            nil,
+	pendingPlugins: make(map[string]*time.Timer),
+	s:              make(chan *framework.Status, 1),
+}
```
sanposhiho, line 195: do we use them?

scheduler/test/util.go
```diff
@@ -71,8 +74,27 @@ func (h *FakeHandle) Parallelizer() (p parallelize.Parallelizer) {
...
+		Spec: v1.PodSpec{NodeName: NodeSmall.Name},
+	}
+
+	waitingPod := &waitingPod{
+		pod:            pod,
+		pendingPlugins: make(map[string]*time.Timer),
+	}
+
+	h.GetWaitingPodValue = waitingPod
+	return waitingPod
```
sanposhiho, line 97: Well... This function makes the test a bit hard to understand.

Can we do this instead?

```suggestion
// GetWaitingPod returns PodForHandleTest only when the uid is handle-test.
func (h *FakeHandle) GetWaitingPod(uid types.UID) framework.WaitingPod {
	if uid != types.UID("handle-test") {
		return nil
	}

	waitingPod := &waitingPod{
		pod:            PodForHandleTest,
		pendingPlugins: make(map[string]*time.Timer),
	}

	h.GetWaitingPodValue = waitingPod
	return waitingPod
```

And, at the test case, you have to swap PodSmall and PodForHandleTest.
Then, you should be able to remove `PodForHandlePod` too. (can use `PodForHandleTest` at `makeTestWaitingPod` instead)

guest/internal/mem/mem.go
chansuke: no, i will remove that :bow:

scheduler/test/testdata.go
chansuke: `waitingPod` is called in https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/blob/fddccd7185ba0a7cf30095aa0f3cbbd94a10d820/scheduler/test/util.go#L82 but no need WaitingPod

scheduler/plugin/plugin_test.go
chansuke: expectedUID is redundant, thank you

scheduler/test/util.go
chansuke: I appreciate your feedback. That's simple and clean.

scheduler/test/testdata.go
```diff
@@ -143,6 +145,46 @@ var PodForHandleTest = &v1.Pod{
 	Spec: v1.PodSpec{NodeName: NodeSmall.Name},
 }
 
+var PodForHandlePod = &v1.Pod{
```
sanposhiho, line 148: Do we need it?

scheduler/test/testdata.go
```diff
@@ -143,6 +145,46 @@ var PodForHandleTest = &v1.Pod{
...
+
+func (wp *waitingPod) Allow(pluginName string) {
+	wp.mu.Lock()
+	defer wp.mu.Unlock()
+}
+
+func (wp *waitingPod) Reject(reason string, msg string) {
+	wp.mu.Lock()
+	defer wp.mu.Unlock()
+}
```
sanposhiho, line 186: any reason you implemented it here, not at scheduler/test/util.go?

scheduler/plugin/plugin_test.go
```diff
@@ -1807,3 +1886,48 @@ func requireError(t *testing.T, err error, expectedError string) {
 		t.Fatalf("unexpected error: want %v, have %v", want, have)
 	}
 }
+
+type waitingPod struct {
```
sanposhiho, line 1890: why do we have to implement `waitingPod` here as well in addition to scheduler/test/testdata.go?

scheduler/test/testdata.go
chansuke: no, we don't need it.

scheduler/test/testdata.go
chansuke: i will move this into `scheduler/test/util.go`

scheduler/plugin/plugin_test.go
```diff
@@ -1807,3 +1886,7 @@ func requireError(t *testing.T, err error, expectedError string) {
 		t.Fatalf("unexpected error: want %v, have %v", want, have)
 	}
 }
+
+func makeTestWaitingPod(pod *v1.Pod, plugins map[string]*time.Timer) framework.WaitingPod {
+	return util.NewWaitingPod(pod, plugins)
+}
```
sanposhiho, line 1892: Do we need this function. It's just an alias.

sanposhiho (commented):
oth

scheduler/plugin/plugin_test.go
chansuke: i will remove it

sanposhiho (approved):
/lgtm
/approve

Awesome!

Issue #101, opened 2024-01-29
Grow examples 


sanposhiho:
We have a few examples only for very basic usecases. On the other hand, we have almost all the features the scheduling framework has. The scheduling framework is well created to cover wider usecases and so does our wasm extension. 

Growing examples contributes to them:
- Growing the variety of examples would be a great showcase for users who have different usecases.
- If we implement some existing plugins in [in-tree plugins](https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/framework/plugins), we can have a benchmark comparison between Golang plugin and wasm plugin via [scheduler_perf](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/tree/main/internal/e2e/scheduler_perf).
- Cover multiple practical scenarios in [e2e tests](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/tree/main/internal/e2e/scheduler).

dejanzele:
Hello all,

Thanks for creating such an interesting project!

I am a fellow Kubernetes contributor and I am interested in getting involved with this project.

I am interested in taking on one of the plugins from the https://github.com/kubernetes-sigs/scheduler-plugins and porting it to WASM.

Do you have any preference on which one?

sanposhiho:
Hi @dejanzele. Thanks, feel free to create a PR and then ping me, @Gekko0114 and @utam0k.

Speaking of which, instead of sigs/scheduler-plugins, [in-tree plugins](https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/framework/plugins) should be priority now.
I suggest [nodeunschedulable](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/nodeunschedulable/node_unschedulable.go) as a starting point, which I believe we can implement it without any additional feature to the current SDK.

unvavo:
Hi @sanposhiho. 
I would like to work on this issue. Do you have any recommendations on which in-tree plugins to work on?

Issue #118, opened 2024-08-23
Explore Go SDK


sanposhiho:
Looks like `go:wasmimport` is now available at gotip: https://github.com/golang/go/issues/65199

We can explore Go SDK, as an alternative to TinyGo SDK. We can take a benchmark etc and compare two SDKs after all.

Also, moving to Go SDK will allow us to use k8s package directly, which would make it easier to port more in-tree plugins.

/kind feature

musaprg:
@sanposhiho Hi. I'm interested in this topic. Can I try to make this happen?

Recently I've been working on [OTelWasm](https://github.com/otelwasm/otelwasm/tree/main). It's mostly based on the idea and code living in this project, but I wrote guest SDK with latest Go1.24 instead of TinyGo. I believe I could move this forward based on our experiments done as part of otelwasm project.

musaprg:
Thanks. I'll look into the existing PR later.

/assign

PR #123, opened 2024-11-05
[WIP] feat: port the NodeUnschedulable in-tree plugin to wasm

dejanzele:
#### What type of PR is this?

/kind feature

Ports the in-tree NodeUnschedulable plugin

Relates https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/101

This is a work in progress, I have made it work locally, but this is not finished yet. I have created this PR to get early feedback as I am not sure some things need to be handled the way I handled them.

I need to add more tests, will do that in the next couple of days.

```release-note
NONE
```

```benchstat

```


Makefile
```diff
@@ -5,13 +5,16 @@ golangci_lint := github.com/golangci/golangci-lint/cmd/golangci-lint@v1.53.2
...
 
+examples/nodeunschedulable/main.wasm: examples/nodeunschedulable/main.go
+	@(cd $(@D); tinygo build -o main.wasm -gc=custom -tags=custommalloc -scheduler=none --no-debug -target=wasi .)
+
 %/main.wasm: %/main.go
 	@(cd $(@D); tinygo build -o main.wasm -scheduler=none --no-debug -target=wasi .)
 
 .PHONY: build-tinygo
-build-tinygo: examples/nodenumber/main.wasm examples/advanced/main.wasm examples/imagelocality/main.wasm guest/testdata/cyclestate/main.wasm guest/testdata/filter/main.wasm guest/testdata/score/main.wasm \
+build-tinygo: examples/nodenumber/main.wasm examples/advanced/main.wasm examples/imagelocality/main.wasm examples/nodeunschedulable/main.wasm guest/testdata/cyclestate/main.wasm guest/testdata/filter/main.wasm guest/testdata/score/main.wasm \
```
dejanzele, line 15: I do not understand why the `internal/e2e/scheduler_perf/wasm/nodeunschedulable` can be compiled successfully with the `%/main.wasm` but the `examples/nodeunschedulable` one which is almost identical fails and needs the `-gc=custom -tags=custommalloc` flags.

guest/api/framework.go
```diff
@@ -1,5 +1,16 @@
 package api
 
+type QueueingHint int
```
dejanzele, line 3: This refers to https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/types.go#L256-L263. Where should this be placed?

guest/api/helper/helpers.go
```diff
@@ -0,0 +1,16 @@
+package helper
+
+import (
+	"sigs.k8s.io/kube-scheduler-wasm-extension/guest/api"
+	protoapi "sigs.k8s.io/kube-scheduler-wasm-extension/kubernetes/proto/api"
+)
+
+// TolerationsTolerateTaint checks if taint is tolerated by any of the tolerations.
+func TolerationsTolerateTaint(tolerations []*protoapi.Toleration, taint *protoapi.Taint) bool {
```
dejanzele, line 9: This refers to this function https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/toleration.go#L38-L57.
As the `kubernetes` folder is sparsely checked out and I guess should not be modified, is there a better approach how to use this function?

guest/api/tolerations.go
```diff
@@ -0,0 +1,46 @@
...
+// ToleratesTaint checks if the toleration tolerates the taint.
+// The matching follows the rules below:
+//
+//  1. Empty toleration.effect means to match all taint effects,
+//     otherwise taint effect must equal to toleration.effect.
+//  2. If toleration.operator is 'Exists', it means to match all taint values.
+//  3. Empty toleration.key means to match all taint keys.
+//     If toleration.key is empty, toleration.operator must be 'Exists';
+//     this combination means to match all taint values and all taint keys.
+func ToleratesTaint(toleration *v1.Toleration, taint *v1.Taint) bool {
```
dejanzele, line 27: This refers to this function https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/toleration.go#L38-L57.
As the kubernetes folder is sparsely checked out and I guess should not be modified, is there a better approach how to use this function?

internal/e2e/scheduler_perf/config/sched-config-all-enabled.yaml
```diff
@@ -5,11 +5,11 @@ profiles:
       multiPoint:
         enabled:
           - name: wasm
-          - name: NodeNumber
+          - name: NodeUnschedulable
```
dejanzele, line 8: This was edited so I can test does it work.
I'll fix it with a better config.

dejanzele:
Here is an AI generated summary of the benchmark report (should be taken with a bit of reservation):

# Report on Scheduler Performance: Go In-tree vs. WASM Plugin

## Scheduling Throughput (pods/s):
- **Default Go In-tree**: Generally has a higher throughput (e.g., ~183.8 pods/s).
- **WASM Plugin**: Lower scheduling throughput compared to the Go in-tree version (e.g., ~55.6 pods/s for some configurations).
- **Analysis**: The Go in-tree scheduler demonstrates better throughput, indicating it can schedule more pods per second. This is a key metric for scenarios requiring high scalability and rapid scheduling.

## Scheduling Attempt Duration (ms):
- **Go In-tree**: Shows shorter average and percentile durations (e.g., ~20.97 ms for Average, 20.8 ms for Perc50).
- **WASM Plugin**: Exhibits longer attempt durations (e.g., ~25.9 ms for Average and higher 99th percentile values like 563.2 ms).
- **Analysis**: The Go in-tree scheduler performs better in terms of scheduling attempt duration, suggesting it is quicker at attempting to schedule individual pods compared to the WASM-based plugin.

## Pod Scheduling Duration (ms):
- **Go In-tree**: More consistent with shorter average pod scheduling times (e.g., ~2430.5 ms for Average).
- **WASM Plugin**: Longer average scheduling times with higher percentile spikes (e.g., ~8030.8 ms Average and significant jumps at Perc90 and Perc99 levels).
- **Analysis**: The Go in-tree scheduler provides more stable and quicker scheduling times, indicating a more predictable performance under load. The WASM plugin shows greater variability and higher latency, particularly under stress.

## Extension Point Durations (ms):
- **PreFilter, Filter, Score, Bind, etc.**:
  - **Go In-tree**: Generally faster across extension points (e.g., averages in the sub-millisecond range).
  - **WASM Plugin**: Shows more variability and higher values in specific extension points (e.g., Filter averaging at ~15.98 ms compared to Go's sub-millisecond figures).
- **Analysis**: The Go scheduler demonstrates better performance across the different extension points of the scheduling process, which contributes to overall scheduling efficiency.

## High Percentile Analysis (Perc99):
- **Go In-tree**: Tends to have lower 99th percentile values, indicating more consistent performance even under heavy load.
- **WASM Plugin**: Often displays higher 99th percentile times, suggesting potential bottlenecks or slower edge-case performance.
- **Implication**: High 99th percentile values in the WASM plugin imply that, in scenarios with peak demand, performance can degrade significantly, leading to higher scheduling latency.

sanposhiho:
Thanks for the PR @dejanzele. 
FYI though, I'm away for Kubecon and most likely cannot visit here until the week after next. Pre-apologize for that!

sanposhiho:
@Gekko0114 @utam0k can you take a look as well?

guest/api/framework.go
sanposhiho: Why do we need this in the first place? We don't support QueueingHint in the extension yet.

guest/api/helper/helpers.go
sanposhiho: You should put it (along with `ToleratesTaint(...)`) inside examples/nodeunschedulable.
I don't think we should maintain this function as the part of the guest SDK.

guest/api/tolerations.go
sanposhiho: Yes, kubernetes folder is only for generating the object files. Also, we cannot import the things from kubernetes because of the limitation coming from TinyGo.
So, yes, it's the only option that we almost do a copy like you do here, at least until we move to [Golang SDK](https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/118).

internal/e2e/scheduler_perf/wasm/nodeunschedulable/main.go
sanposhiho, line 1: Why do we need this? I mean, can we just use the wasm built from examples/nodeunschedulable?

internal/e2e/scheduler_perf/config/sched-config-all-enabled.yaml
sanposhiho, line 1: Please revert all the changes in scheduler-perf test cases.
If you want to include the nodeunschedulable plugin in the scheduler-perf, it's better to create a new test case, not replacing the existing test cases to use nodeunschedulable instead of nodenumber.
nodenumber is the simplest plugin which is good to measure the overhead of wasm extension.

dejanzele:
Hi,

A lot of things happened after KubeCon, I plan to resume working on this in a day or two.

PR #126, opened 2024-11-10
Upgrade `golangci-lint` to 1.61.0

jgiannuzzi:
The previous version of `golangci-lint` (1.53.2) crashed under recent versions of Go.

#### What type of PR is this?

/kind cleanup

The `lint` make target currently fails with recent versions of Go (e.g. 1.23.3) because the version of `golangci-lint` that we use crashes at runtime.
Updating to 1.61.0 solves this.

I also had to update the version of Go used in the CI workflows to get them to pass.
I thus updated to the latest version, and did the same for TinyGo.

The latest TinyGo now has multiple `wasi` targets, and the main one inherits from `wasip1`. I had to change the way we generate the test target to deal with this inheritance.

```release-note
NONE
```

```benchstat
N/A
```

### DIFF
--- a/.github/workflows/commit.yaml
+++ b/.github/workflows/commit.yaml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         go-version:  # Note: Go only supports 2 versions: https://go.dev/doc/devel/release#policy
-          - "1.20"
+          - "1.23"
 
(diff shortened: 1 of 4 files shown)


sanposhiho (approved):
/lgtm
/approve

PR #127, opened 2024-11-10
Improve benchmark performance of nodenumber wasm plugin

jgiannuzzi:
#### What type of PR is this?

/kind cleanup

The `nodenumber` WASM plugin used in the scheduler_perf benchmark doesn't use advanced techniques like the custom GC and advertising only the implemented endpoints.
By adopting these techniques and rebuilding the WASM module, the benchmark suggests that the WASM plugin performance is between the native code and the extender, whilst it was slower than the extender before this change.

I have also changed the location of the WASM module in the `AllEnabled` profile, as it was pointing to the `nodenumber` _example_ rather than the one we build for the benchmark.

```release-note
NONE
```

```benchstat
N/A
```

Running the `scheduler_perf` benchmark 3 times on the same machine and taking the average of the results for scheduling 10000 pods, this results in a ~10% increase, going from 750 pods/s to 833 pods/s.


sanposhiho (approved):
/lgtm
/approve

PR #131, opened 2024-11-13
feat: add support for GetLabels and GetAnnotations in metadata

dejanzele:
#### What type of PR is this?

/kind feature

This PR adds support for `GetLabels` and `GetAnnotations` in the `Metadata` interface so labels and annotations can be fetched for pods and nodes.

This will be required when porting in-tree plugins as some of them need access to labels and annotations.

Fixes #

```release-note
NONE
```

```benchstat

```


guest/internal/proto/proto.go
```diff
@@ -54,12 +54,34 @@ func GetResourceVersion[O object](o O) string {
 	return ""
 }
 
+func GetLabels[O object](o O) map[string]string {
+	if md := o.GetMetadata(); md != nil {
```
sanposhiho, line 58:
```suggestion
	if md := o.GetLabels(); md != nil {
```

dejanzele:
```
type object interface {
	GetMetadata() *meta.ObjectMeta
}
```

Are you saying to add `GetLabels` and `GetAnnotations` methods to the `object` interface?

sanposhiho (approved):
/lgtm
/approve

PR #132, opened 2024-11-14
fix nodenumber plugin to use Pod name instead of pod.spec.nodeName

YZ775:
#### What type of PR is this?
/kind bug

The nodeNumber plugin is intended to scores higher if the node name and pod name suffixes match.
But in implementation, it uses Pod's .`Spec.nodeName` not a `.metadata.Name`.
So PreScore allways return nothing, and therefore allways `match=true` in Score.

This PR fix this problem

Fixes #

```release-note
change nodeNumber plugin to follow specification.
```

[after.txt](https://github.com/user-attachments/files/17759131/after.txt)
[before.txt](https://github.com/user-attachments/files/17759122/before.txt)
```benchstat
❯ benchstat before.txt after.txt
goos: darwin
goarch: arm64
cpu: Apple M1
                                      │  before.txt  │              after.txt              │
                                      │    sec/op    │    sec/op     vs base               │
Example_NodeNumber/Simple/New-8         497.5m ± 14%   506.5m ±  1%        ~ (p=0.394 n=6)
Example_NodeNumber/Simple/Run-8         177.8µ ± 30%   169.1µ ± 10%        ~ (p=0.310 n=6)
Example_NodeNumber/Simple_Log/New-8     531.7m ± 19%   501.5m ±  4%        ~ (p=0.310 n=6)
Example_NodeNumber/Simple_Log/Run-8     206.7µ ± 17%   213.9µ ± 19%        ~ (p=0.310 n=6)
Example_NodeNumber/Advanced/New-8        1.081 ± 12%    1.076 ±  5%        ~ (p=1.000 n=6)
Example_NodeNumber/Advanced/Run-8       62.07µ ±  4%   69.25µ ±  3%  +11.56% (p=0.002 n=6)
Example_NodeNumber/Advanced_Log/New-8    1.084 ± 80%    1.066 ±  4%        ~ (p=0.937 n=6)
Example_NodeNumber/Advanced_Log/Run-8   76.83µ ± 17%   79.26µ ±  2%        ~ (p=0.394 n=6)
geomean                                 9.265m         9.338m         +0.78%
```

### DIFF
--- a/examples/nodenumber/main.go
+++ b/examples/nodenumber/main.go
@@ -96,7 +96,7 @@ func (pl *NodeNumber) PreScore(state api.CycleState, pod proto.Pod, _ api.NodeIn
 
 	klog.InfoS("execute PreScore on NodeNumber plugin", "pod", klog.KObj(pod))
 
-	podnum, ok := lastNumber(pod.Spec().GetNodeName())
+	podnum, ok := lastNumber(pod.GetName())
 	if !ok {
 		recorder.Eventf(pod, nil, "PreScore", "not match lastNumber", "Skip", "")
 		return nil // return success even if its suffix is non-number.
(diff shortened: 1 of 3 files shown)


sanposhiho (approved):
/lgtm
/approve

PR #141, opened 2025-06-11
Update to Kubernetes 1.32 and Go 1.23

utam0k:
#### What type of PR is this?

/kind cleanup
/sig scheduling

Uupdate to Kubernetes 1.32 and Go 1.23

Fixes https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/issues/140

```release-note
Uupdate to Kubernetes 1.32 and Go 1.23
```

```benchstat

```


sanposhiho:
Why not jumping to Go 1.24? I think we can.

utam0k:
@sanposhiho I tried Go 1.24, but TinyGo 0.35+ requires Go 1.24, which doesn't support protobuf 1.30. More precisely, it doesn't support `reflect.SliceOf` I would like to give up Go 1.24 for this PR.

PR #143, opened 2025-11-10
ci: make sure the Wasm files are latest

saza-ku:
#### What type of PR is this?
/kind bug

The current ci action does `make testdata` to check the Wasm files are latest, but it reports nothing even if the guest's codes are changed without updating the Wasm files.

`make` checks the timestamp of Go files and Wasm files. But timestamps of all the files are the same in the ci environment, because they are set to time when the repository is checked out. So `make` always thinks Wasm files are latest in ci.

To fix the issue, this PR makes sure the Wasm files are latest by cleaning up the Wasm files before doing `make testdata`.

Fixes #

```release-note

```

```benchstat

```

### DIFF
--- a/.github/workflows/testdata.yaml
+++ b/.github/workflows/testdata.yaml
@@ -52,6 +52,9 @@ jobs:
           wabt_url=https://github.com/WebAssembly/wabt/releases/download/${wabt_version}/wabt-${wabt_version}-ubuntu-20.04.tar.gz
           curl -sSL ${wabt_url} | tar --strip-components 2 -C /usr/local/bin -xzf - wabt-${wabt_version}/bin/wat2wasm
 
+      - name: "clean testdata"
+        run: make clean-testdata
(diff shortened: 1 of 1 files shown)


.github/workflows/commit.yaml
```diff
@@ -6,7 +6,8 @@ name: "build"
 
 on:
   push:  # We run tests on non-tagged pushes to main
-    tags: ''
```
saza-ku: > Invalid workflow file: .github/workflows/testdata.yaml#L1
(Line: 9, Col: 11): Unexpected value ''


https://github.com/kubernetes-sigs/kube-scheduler-wasm-extension/actions/runs/19224097301

utam0k: 👀

utam0k (commented):
/lgtm
from my perspective, lgtm

